PR #20, opened 2025-04-23
Reka Flash 3 vs padding

gakada:
I'm not sure how this might interact with other models, but it fixes [Reka Flash 3](https://huggingface.co/RekaAI/reka-flash-3) quantization. Otherwise, there are attempts to add and multiply matrices of incompatible sizes, because the model has both output and input sizes that are not divisible by 128, but only the output is padded.

reka

Here is HumanEval for [[1]](https://huggingface.co/lucyknada/RekaAI_reka-flash-3-exl2) and [[2]](https://huggingface.co/isogen/reka-flash-3-exl3-3bpw) (eager with no reasoning block, sampling with reasoning and full precision is supposed to be 96.3).

| quant     | KV   | GB   | prompt | pass@1 (1K, 2K) |
| --------- | ---- | ---- | ------ | --------------- |
| exl2 3bpw | Q4   | 10.7 | raw    | 67.1, 71.3      |
| exl2 3bpw | Q8   | 11.8 | raw    | 72.0, 72.6      |
| exl2 3bpw | FP16 | 12.9 | raw    | 70.1, 69.5      |
| exl2 3bpw | Q4   | 10.7 | reka   | 77.4, 77.4      |
| exl2 3bpw | Q8   | 11.8 | reka   | 75.6, 78.7      |
| exl2 3bpw | FP16 | 12.9 | reka   | 77.4, 73.2      |
| exl2 4bpw | Q4   | 12.2 | raw    | 86.0, 84.8      |
| exl2 4bpw | Q8   | 13.2 | raw    | 85.4, 88.4      |
| exl2 4bpw | FP16 | 15.4 | raw    | 87.2, 87.2      |
| exl2 4bpw | Q4   | 12.2 | reka   | 84.8, 86.0      |
| exl2 4bpw | Q8   | 13.2 | reka   | 86.0, 89.6      |
| exl2 4bpw | FP16 | 15.4 | reka   | 86.0, 85.4      |
| exl3 3bpw | Q4   | 8.8  | raw    | 84.8, 86.0      |
| exl3 3bpw | Q8   | 9.0  | raw    | 87.8, 89.0      |
| exl3 3bpw | FP16 | 9.6  | raw    | 88.4, 89.0      |
| exl3 3bpw | Q4   | 8.8  | reka   | 87.2, 87.8      |
| exl3 3bpw | Q8   | 9.0  | reka   | 88.4, 90.2      |
| exl3 3bpw | FP16 | 9.6  | reka   | 90.2, **90.9**  |
| exl3 4bpw | Q4   | 11.3 | raw    | 89.0, 88.4      |
| exl3 4bpw | Q8   | 11.5 | raw    | 87.2, 88.4      |
| exl3 4bpw | FP16 | 12.3 | raw    | 87.2, 87.2      |
| exl3 4bpw | Q4   | 11.3 | reka   | 86.6, 85.4      |
| exl3 4bpw | Q8   | 11.5 | reka   | 84.1, 86.6      |
| exl3 4bpw | FP16 | 12.3 | reka   | 86.0, 86.6      |

### DIFF
--- a/eval/humaneval.py
+++ b/eval/humaneval.py
@@ -49,6 +49,13 @@ prompt_formats = {
         "<start_of_turn>model\n"
         "```python\n{{problem}}",
         "    "
+    ),
+    "reka": (
+        "human: Complete the following Python function."
+        " Provide your reasoning in comments, but be concise and don't second-guess."
+        "\n\n{{problem}}"
+        " <sep> assistant: ```python\n{{problem}}",
+        "    "
     )
 }
 
--- a/exllamav3/modules/linear.py
+++ b/exllamav3/modules/linear.py
@@ -38,7 +38,8 @@ class Linear(Module):
         super().__init__(config, key, qmap)
 
         self.alt_key = alt_key
-        self.in_features = in_features
+        self.in_features_unpadded = in_features
+        self.in_features = (in_features + pad_to - 1) // pad_to * pad_to
         self.out_features_unpadded = out_features
         self.out_features = (out_features + pad_to - 1) // pad_to * pad_to
         self.full_in_features = full_in_features
@@ -59,11 +60,14 @@ class Linear(Module):
 
 
     def pad_out(self, w: torch.Tensor | None) -> torch.Tensor | None:
-        if w is None or self.out_features == self.out_features_unpadded:
+        if w is None or self.out_features == self.out_features_unpadded and self.in_features == self.in_features_unpadded:
             return w
         if w.dim() == 2:
             padded = torch.zeros((self.out_features, self.in_features), dtype = w.dtype, device = w.device)
-            padded[:w.shape[0], :] = w
+            if self.out_features != self.out_features_unpadded:
+                padded[:w.shape[0], :] = w
+            elif self.in_features != self.in_features_unpadded:
+                padded[:, :w.shape[1]] = w
         else:
             assert w.dim() == 1
             padded = torch.zeros((self.out_features,), dtype = w.dtype, device = w.device)


turboderp:
This seems okay, though it should definitely be done with some care. For MLP layers the down projection can be trivially padded with zeros before quantizing, since the extra input channels will never contribute anything to the output. But attention (and down the line convolutions and other stuff) might break in subtle and confusing ways if both input and output channels are padded automatically.

I'll give this a closer look tomorrow. Looks like a very interesting model, too.

gakada:
By the way, right now it will still error on simultaneous padding.

```py
import torch

def test(in_features, out_features, pad_to):
    w = torch.zeros((out_features, in_features))

    in_features_unpadded = in_features
    in_features = (in_features + pad_to - 1) // pad_to * pad_to
    out_features_unpadded = out_features
    out_features = (out_features + pad_to - 1) // pad_to * pad_to

    padded = torch.zeros((out_features, in_features))
    if out_features != out_features_unpadded:
      padded[:w.shape[0], :] = w
    elif in_features != in_features_unpadded:
      padded[:, :w.shape[1]] = w

    print(w.shape)
    print(padded.shape)

test(64, 128, 8) # ok
test(64 + 4, 128, 8) # ok
test(64, 128 + 4, 8) # ok
test(64 + 4, 128 + 4, 8) # error
```

turboderp:
In that case `w` doesn't match the shape of either `padded[: w.shape[0], :]` or `padded[:, : w.shape[1]]`. You'd want to pad both at the same time, something like:

```py
    if (in_features, out_features) != (in_features_unpadded, out_features_unpadded):
        padded = torch.zeros((out_features, in_features))
        padded[: out_features_unpadded, :in_features_unpadded] = w
```

gakada:
Yeah, I just mean that maybe it is better to leave it as an error, if there are indeed any subtleties. Until there is a model that needs it, so it can be actually tested.

turboderp:
Ah, I see, so restrict it to only padding one dimension. Yeah, that makes sense.

turboderp:
Merged it now, but only enabled for MLP down layer to avoid accidentally breaking anything else. Also added reasoning template for the chat script. Very cerebral model. :)

image

PR #25, opened 2025-04-27
Language layers for Mistral Small 3.1 and InternVL3

gakada:
They are the same Mistral/Qwen2/Llama layers, so quantization and text inference is already possible.

### DIFF
--- a/exllamav3/exllamav3_ext/bindings.cpp
+++ b/exllamav3/exllamav3_ext/bindings.cpp
@@ -2,6 +2,8 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
+#include <cuda_fp16.h>
+
 #include "stloader.h"
 #include "hadamard.h"
 
@@ -16,6 +18,7 @@
 #include "quant/reconstruct.cuh"
 #include "quant/hadamard.cuh"
 #include "quant/exl3_gemm.cuh"
+#include "quant/exl3_kernel_map.cuh"
 #include "quant/util.cuh"
 
 #include "generator/strings.h"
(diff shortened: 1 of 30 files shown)


turboderp:
I've been holding off a little on Mistral 3.1 because I want to work out a sane way to support component models. Quantizing this way would give you Mistral 3.1 quants all over HF that won't work with the (upcoming, soon™ ) multimodal features. Patching in support this way when it's "easy" has become a real headache in ExLlamaV2, and I'm trying to be a little more deliberate about it this time around. I'm still not sure exactly how it should look, but I plan to start architecting the multimodal support next week and I'll have a clearer picture then.

As for dynamic RoPE, it generally can't work with the current cache implementation since keys are stored after applying embeddings, so the cached keys wouldn't stay valid for a growing context. TBH I never fully understood how it's meant to work in Transformers either, since the usual caching looks like:

```py
        query_states = self.q_proj(hidden_states).view(hidden_shape).transpose(1, 2)
        key_states = self.k_proj(hidden_states).view(hidden_shape).transpose(1, 2)
        value_states = self.v_proj(hidden_states).view(hidden_shape).transpose(1, 2)

        cos, sin = position_embeddings
        query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)

        if past_key_value is not None:
            # sin and cos are specific to RoPE models; cache_position needed for the static cache
            cache_kwargs = {"sin": sin, "cos": cos, "cache_position": cache_position}
            key_states, value_states = past_key_value.update(key_states, value_states, self.layer_idx, cache_kwargs)
```

Then `@dynamic_rope_update` recomputes embeddings every time a token is added, but the cached keys are never updated to the new frequencies (and technically can't be without running the entire context through the forward pass, on every token.) I can see it working if keys are cached prior to embedding and RoPE is done on-the-fly, but there's a performance penalty for that, too, and support for fusing it into the attention step is patchy with both flash-attn and FlashInfer (which I'm hoping to migrate to when it supports more head dimensions.) So I've kind of been deliberately ignoring dynamic RoPE until it starts making sense to me. (:

gakada:
They are basically `MistralForCausalLM` and `Qwen2ForCausalLM` quants, so should still work, maybe with fixed `config.json` and `*.safetensors` (no requantization). Though I can also just convert to `MistralForCausalLM` and `Qwen2ForCausalLM` first and quantize that, so I removed the hack :ok_hand:

The dynamic RoPE in InternVL3 only happens past 32K context, at which point those models aren't very good anyway, so it is probably fine to have it as default. But it should be there because it can be in `config.json`.

gakada:
When multimodality is implemented, do you think it is worth supporting multimodal → language-only conversions? Like InternVL3 38B is 6B of ViT, so some significant memory can be saved, if vision isn't needed. Alternatively, those layers can be disabled during inference, but then you'd still download and store them.

Or just do such conversions externally to exllamav3?

If implemented here, it can be something like

```py
{
  "Mistral3ForConditionalGeneration": ("MistralForCausalLM", "language_model."),
  "InternVLChatModel": ("Qwen2ForCausalLM", "language_model."),
  # ...
}

# use this map when convert.py is asked for a language-only conversion
# allow safetensors to load layers with a prefix
# everything will see prefixless layer names and save them as such
# then convert config.json
```

gakada:
@turboderp right now gemma-3-1b also won't run because it is a language-only model with `Gemma3ForCausalLM`.

PR #30, opened 2025-04-30
fix: don't add min_p and top_p samplers to stack if falsy

randoentity:
Got AssertionError because the samplers (min_p, top_p) were not allowed to be 0.0.
I don't know if I should set them to 1.0 instead to turn them off.
Because of these unclear (to me; severe memory issues) values for "off" I did not set this for the other samplers (they did not have the asserts).

### DIFF
--- a/exllamav3/generator/sampler/presets.py
+++ b/exllamav3/generator/sampler/presets.py
@@ -109,11 +109,12 @@ class ComboSampler(CustomSampler):
         else:
             stack += [
                 SS_Temperature(temperature if not temp_last else 1.0),
-                SS_MinP(min_p),
+                SS_MinP(min_p) if min_p else None,
                 SS_TopK(top_k),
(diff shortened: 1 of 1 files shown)


turboderp:
I actually just had the wrong assertion on min-P, which is fixed now. A value of 0.0 means disabled (mask out all tokens whose probability is less than zero times the top token's probability).

For top-P the disabled value is 1.0 (keep 100% of the probability mass.)

I'd like to keep these somewhat strict since the idea is to add a pass that optimizes an arbitrary sampler stack later on, and that will be a lot easier if the input arguments make sense across their whole respective ranges and well-defined at the extremes. 

For instance for top-P, the lower the argument is, the closer it is to greedy sampling. It makes sense that top_p = 0.0 gives roughly the same result as top_p = 0.001.

PR #35, opened 2025-05-11
humaneval.py: fix top_k type, remove rep_p, add qwen3

gakada:
Missed `top_k` being `int`, so it doesn't work as an index like that.

### DIFF
--- a/eval/humaneval.py
+++ b/eval/humaneval.py
@@ -66,6 +66,15 @@ prompt_formats = {
         "Sure! Here is how you might implement the function:\n\n```python\n{{problem}}",
         "    "
     ),
+    "qwen3": (
+        "<|im_start|>system\n"
+        "You are a helpful AI coding assistant.<|im_end|>\n"
+        "<|im_start|>user\n"
+        "Complete the following Python function:\n\n{{problem}}<|im_end|>\n"
+        "<|im_start|>assistant\n"
+        "<think>\n\n</think>\n\nSure! Here is how you might implement the function:\n\n```python\n{{problem}}",
+        "    "
+    ),
     "deepseek": (
         "You are a helpful AI coding assistant.\n"
         "<｜User｜>Complete the following Python function:\n\n{{problem}}"
@@ -99,7 +108,6 @@ def main(args):
         tokenizer = tokenizer
     )
     sampler = ComboSampler(
-        rep_p = args.rep_p,
         temperature = args.temperature,
         min_p = args.min_p,
         top_k = args.top_k,
@@ -189,10 +197,9 @@ if __name__ == "__main__":
     parser.add_argument("-e", "--eval", action = "store_true", help = "Run evaluation script on output file after sampling")
     parser.add_argument("-temp", "--temperature", type = float, help = "Sampling temperature (0 for greedy), default: 0.6", default = 0.6)
     parser.add_argument("-minp", "--min_p", type = float, help = "Min-p sampling, default: 0.0 (disabled)", default = 0.0)
-    parser.add_argument("-topk", "--top_k", type = float, help = "Top-k sampling, default: 0.0 (disabled)", default = 0.0)
+    parser.add_argument("-topk", "--top_k", type = int, help = "Top-k sampling, default: 0 (disabled)", default = 0)
     parser.add_argument("-topp", "--top_p", type = float, help = "Top-p sampling, default: 0.6", default = 0.6)
     parser.add_argument("-templast", "--temp_last", action = "store_true", help = "Use temperature last")
-    parser.add_argument("-repp", "--rep_p", type = float, help = "Repetition penalty, default: 1.0 (disabled)", default = 1.0)
     parser.add_argument("--max_tokens", type = int, default = 768, help = "Max number of tokens for each completion")
     _args = parser.parse_args()
     main(_args)


gakada:
For `rep_p`, all of those break the generation here, so it gets 0%.

```py
rep_p = 1.01,
rep_p = 1.01, rep_sustain_range = 10,
rep_p = 1.01, rep_sustain_range = 10, rep_decay_range = 10,
```

gakada:
`chatml` vs `qwen3` is ~30% vs ~50% on Qwen3-8B 4bpw and ~35% vs ~40% on 8bpw-h8 (and still many strange repetitive completions).

turboderp:
Yep, I checked, and there was a bug in the penalty kernel. Masking out a token and then applying any penalty to it before the falloff range would result in `lerp(-inf, -inf, 1)`, which should yield `-inf` but is evalutated as `(-inf) * 1.0f + (-inf) * 0.0f = NaN`. In this case it was `<|im_end|>` being prohibited because of the `min_new_tokens` argument. Fix will appear in dev branch shortly.

gakada:
I didn't see any difference with `/no_think`, trying to generate functions bodies like that, with or without `/no_think`, can produce many placeholders (like `# Insert implementation here`) and bad repetitions (just repeating the same lines). `enable_thinking=False` seems better, i.e. starting the reply with `<think>\n\n</think>\n\n`, then it is only bad repetitions (less for Qwen3-14B). So yeah, probably just the non-thinking mode not working very well, like reka-flash-3 still manages to think in comments and isn't doing any placeholders or repetitions when trying to generate code before thinking blocks (but it will keep rambling in comments instead). I'm getting

| Model                  | Q4   | Q6   | Q8   | FP16 |
| ---------------------- | ---- | ---- | ---- | ---- |
| Qwen3-8B-exl3-4bpw     | 44.5 | 45.7 | 47.6 | 48.2 |
| Qwen3-8B-exl3-6bpw     | 37.8 | 43.3 | 46.3 | 45.1 |
| Qwen3-8B-exl3-8bpw-h8  | 29.9 | 37.8 | 37.8 | 39.0 |
| Qwen3-14B-exl3-4bpw    | 79.3 | 80.5 | 80.5 | 79.3 |
| Qwen3-14B-exl3-6bpw    | 77.4 | 78.7 | 79.3 | 79.3 |
| reka-flash-3-exl3-3bpw | 87.8 | 87.2 | 90.2 | 90.9 |

And the failures look like this

[Qwen3-8B-exl3-4bpw-q8.md](https://github.com/user-attachments/files/20150619/Qwen3-8B-exl3-4bpw-q8.md)
[Qwen3-8B-exl3-8bpw-h8-q8.md](https://github.com/user-attachments/files/20150618/Qwen3-8B-exl3-8bpw-h8-q8.md)
[Qwen3-14B-exl3-4bpw-q8.md](https://github.com/user-attachments/files/20150620/Qwen3-14B-exl3-4bpw-q8.md)
[reka-flash-3-exl3-3bpw-q8.md](https://github.com/user-attachments/files/20150621/reka-flash-3-exl3-3bpw-q8.md)

turboderp:
You should be able to test with repetition penalties now in the dev branch. Won't fix the thinking-in-comments exploit that Reka-Flash seems to have discovered, but it might help Qwen3-8B get back on track in a few cases where it gets stuck in a loop.

At the end of the day instruct models (and especially reasoning models) are just going to be harder to evaluate this way. HumanEval is also an old test by now and the problems presumably appear in all the pretraining datasets already. So the absolute score doesn't mean anything anymore. I mostly added it as an easy way to compare different quant levels of the same model against one another, and even that doesn't really seem to work anymore when more precise quants can get significantly *lower* scores, somehow.

Here are some results I just did with Qwen3-30B-A3B. Mostly within the margin of error for a single completion per task, but it's remarkable that a model with (presumably) the same training and finetuning data as Qwen3-8B is so much stronger on this particular test, and so much more consistently so:

| Model    | HumanEval pass@1 |
|----------|------------------|
| 2.25 bpw | 88.41%           |
| 3.00 bpw | 89.63%           |
| 4.00 bpw | 92.07%           |
| 5.00 bpw | 93.29%           |
| 6.00 bpw | 92.68%           |
| FP16     | 91.46%           |

And as mentioned, Qwen3-4B also performs much better. It's got to be worth looking into more rigorously. I'd imagine adding a little bit of random noise to the unquantized model could give you a similar effect in a more controlled way.

gakada:
@turboderp with config/tokenizer_config/added_tokens from [here](https://huggingface.co/unsloth/Qwen3-8B/tree/main) and your quant

| Model              | Q4   | Q6   | Q8   | FP16 |
| ------------------ | ---- | ---- | ---- | ---- |
| Qwen3-8B-exl3-4bpw | 86.0 | 85.4 | 86.0 | 87.2 |

turboderp:
Oohh. The only significant difference I can see is that whose files don't define a `bos_token_id` in config.json.

So that makes sense. We've been fighting with this in TabbyAPI for a bit actually, that specifically Qwen3 (but only some versions of it) degrade significantly when there's a BOS token in the context. Don't know why that didn't occur to me here. It's been a real nightmare to try to work out exactly what the default is supposed to be for any given model, and what takes priority when there is conflicting information in the config. It looks like the Unsloth people solved it by just removing it from the config altogether.

I've reworked the prompt formatting a bit in the latest commit on the dev branch, so BOS isn't added by default except for those templates that should have it. And now it seems I'm also getting consistently good results from Qwen3-8B.

gakada:
The original [tokenizer_config.json](https://huggingface.co/Qwen/Qwen3-8B/blob/main/tokenizer_config.json) also has `"add_bos_token": false` and `"bos_token": null` (same as all Qwen3 models), so maybe that should override `bos_token_id` from config.json.

gakada:
@turboderp something else that is strange, on [Qwen3-0.6B](https://huggingface.co/turboderp/Qwen3-0.6B-exl3), `4.0bpw` and `8.0bpw_H8` branches, I'm getting 0% with `-cq 4` and `-cq 5`, but ~40% with `-cq 6`, same for 1.7B, 0% and ~70%, while 4B and up is always ~80% and up

turboderp:
The tokenizer function takes an explicit `add_bos` argument, and that shouldn't be interpreted as "maybe add a BOS token, unless the model's chat template says not to." The tokenizer in general doesn't care about templates, and that function has many other uses besides formatting an entire context for an instruct-tuned model. Arguably the problem with Qwen3 was that it defined a BOS token in the first place when clearly it's not meant to have one. Though it looks like maybe it goes a bit deeper than that, with *some* versions of Qwen3 being pretrained with BOS and some without? Idk.

In any case, the humaneval.py script should be more in line with the ever-evolving non-standard set by HF models :shrug:. It would maybe be more correct to use Transformers to extract the Jinja template from the model itself and always apply that, though there are potential complications with that approach as well.

As for cache quantizations, some models just don't work well if you compress the cache too much. Qwen3-0.6B isn't just a smaller version of Qwen3-4B, it's pretrained from scratch with different parameters, giving a different distribution of outliers in keys/values. It's not strange that it ends up being more sensitive to cache quantization if it tries to pack more information into its attn heads to begin with. You can try experimenting with values like `-cq 6,4` which has the same overall size as `-cq 5` but assigns more bits to the K channel.

PR #45, opened 2025-05-30
Fix q_norm_tensor None check

gakada:
I was trying to load [this](https://huggingface.co/bullerwins/DeepSeek-R1-0528-Qwen3-8B-exl3-8.0bpw) model, it has `rope`, so that check was reached and torch tensors with multiple values can't be used as booleans.


turboderp:
Thank you. (:

PR #74, opened 2025-08-21
Add support for Seed-OSS

Downtown-Case:
Specifically: https://huggingface.co/collections/ByteDance-Seed/seed-oss-68a609f4201e788db05b5dcd

36B dense, 12T pretraining, but what's extremely interesting to me is:

>  Trained with up-to-512K long context natively.

> Research-Friendly: Given that the inclusion of synthetic instruction data in pre-training may affect the post-training research, we released pre-trained models both with and without instruction data, providing the research community with more diverse options.

For the instruct:

> RULER (128K): 94.6

With nothing *odd* like Qwen3's YaRN.

> We recommend sampling with temperature=1.1 and top_p=0.95.

> Users can flexibly specify the model's thinking budget.

***

Transformers support isn't pulled in yet:

https://github.com/Fazziekey/transformers/tree/seed

I looked at the vllm PR. Keeping in mind I am but a end-user here, I can't find any functional difference between it and Qwen2, other than dual chunk attention, which we don't need?

[Qwen2](https://github.com/FoolPlayer/vllm/blob/seed-oss/vllm/model_executor/models/qwen2.py) vs [SeedOss](https://github.com/FoolPlayer/vllm/blob/seed-oss/vllm/model_executor/models/seed_oss.py): https://www.diffchecker.com/mxcVxOL3/

Hence I'm trying out Qwen2's config and a workaround for the specified 'Default' rope scaling type. It's quantizing now.

### DIFF
--- a/exllamav3/architecture/architectures.py
+++ b/exllamav3/architecture/architectures.py
@@ -19,6 +19,7 @@ from .phi3 import Phi3Model
 from .qwen2 import Qwen2Model
 from .qwen3 import Qwen3Model
 from .qwen3_moe import Qwen3MoeModel
+from .seedoss import SeedOssModel
 from .smollm3 import SmolLM3Model
 
 ARCHITECTURES = {
@@ -49,6 +50,7 @@ ARCHITECTURES = {
         Qwen2Model,
         Qwen3Model,
         Qwen3MoeModel,
+        SeedOssModel,
         SmolLM3Model,
     ]
 }
--- a/exllamav3/architecture/seedoss.py
+++ b/exllamav3/architecture/seedoss.py
@@ -1,0 +1,42 @@
+from typing_extensions import override
+from .llama import LlamaConfig, LlamaModel
+
+# Qwen2 is identical to Llama except for bias on Q, K and V projections, but Linear module automatically
+# detects *.bias tensor
+
+class SeedOssConfig(LlamaConfig):
+    arch_string = "SeedOssForCausalLM"
+
+    def __init__(
+        self,
+        directory: str,
+        **kwargs,
+    ):
+        super().__init__(
+            directory,
+            derived_model = {"text": SeedOssModel},
+            **kwargs
+        )
+
+
+class SeedOssModel(LlamaModel):
+    config_class = SeedOssConfig
+
+    def __init__(
+        self,
+        config: SeedOssConfig,
+        **kwargs
+    ):
+        super().__init__(config, **kwargs)
+
+
+    @override
+    def default_chat_prompt(self, prompt: str, system_prompt: str = None) -> str:
+        p = ""
+        if system_prompt:
+            p += f"<seed:bos>system\n"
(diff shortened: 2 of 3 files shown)


turboderp:
:+1: This seems to work. I added a chat template for chat.py that replicates the default systemt prompt from the jinja template. Very interesting reasoning budget and reflections:

```
<seed:think>Got it, let's see. The user wants a Python script that counts to five. First, I need to remember how to loop in Python. Counting to five usually means starting at 1 and going up to 5, inclusive. 

Option 1: Use a for loop with range. Since range is exclusive of the end, range(1,6) would give 1-5.
<seed:cot_budget_reflect>I have used 129 tokens, and there are 383 tokens remaining for use.</seed:cot_budget_reflect>
 Then print each number.

Option 2: Use a while loop, initialize a counter variable at 1, loop while counter <=5, print, then increment.

Both are simple. Let's go with the for loop first because it's more concise for this case. Let me draft that:

for i in range(1, 6):
    print(i)

Wait, that's it?
<seed:cot_budget_reflect>I have used 256 tokens, and there are 256 tokens remaining for use.</seed:cot_budget_reflect>
 Let me check: range(1,6) is 1,2,3,4,5. Yes. Printing each one will count to five. Alternatively, if they want it on one line, but the question doesn't specify, so probably each number on a new line is fine.

Or the while loop version:

count = 1
while count <= 5:
    print(count)
    count += 1

Both work. Let me present both options maybe, explain each.
<seed:cot_budget_reflect>I have used 393 tokens, and there are 119 tokens remaining for use.</seed:cot_budget_reflect>
 The user might be a beginner, so clarity is good. Let's start with the for loop as it's more Pythonic for iterating over a sequence.</seed:think>
```

turboderp:
For reference:

```py
    def default_system_prompt(self, think):
        if think:
            budget_reflections_v05 = {
                0: 0,
                512: 128,
                1024: 256,
                2048: 512,
                4096: 512,
                8192: 1024,
                1e20: 1024
            }
            ns_interval = budget_reflections_v05[min([k for k in budget_reflections_v05 if k >= self.thinking_budget])]
            return (
                f"You are an intelligent assistant with reflective ability. In the process of thinking and reasoning, "
                f"you need to strictly follow the thinking budget, which is {self.thinking_budget}. That is, you need to "
                f"complete your thinking within {self.thinking_budget} tokens and start answering the user's questions. "
                f"You will reflect on your thinking process every {ns_interval} tokens, stating how many tokens have "
                f"been used and how many are left."
            )
        else:
            return (
                "You are an intelligent assistant that can answer questions in one step without the need for reasoning "
                "and thinking, that is, your thinking budget is 0. Next, please skip the thinking process and directly "
                "start answering the user's questions."
            )
```

The default no-thinking prompt ends up wasting a few tokens, but it seems to respect the budget:

```
<seed:think>The current thinking budget is 0, so I will directly start answering the question.</seed:cot_budget_reflect> </seed:think>Hello! How can I help you today?
```

PR #152, opened 2026-02-25
Switch ExLlamaV3 to flashinfer and add MLA/Qwen3.5 support

lesj0610:
## Goal
Complete the ExLlamaV3 runtime migration to `flashinfer`, keep the new backend stable across the existing serving stack, and land the model/architecture work that became necessary during the migration.

## Status
This PR now represents a **completed flashinfer backend migration**.

What remains after this PR is follow-up optimization work, not migration work.

### Backend migration
- Switched the primary attention runtime path to `flashinfer`.
- Kept safe fallbacks where the backend cannot support a given fast path.
- Added runtime stats toggles for hot-path inspection:
  - `EXLLAMAV3_RUNTIME_STATS=1`
  - `EXLLAMAV3_RUNTIME_STATS_INTERVAL=<N>`

### Core runtime optimizations (common path)
- Reuse FlashInfer prefill wrappers/plans across layers.
- Reuse generator batch metadata buffers (`block_index`, `cache_seqlens`, `positions`).
- Avoid `torch.cat(...)` in the single-token decode hot path.
- Reduce non-tensor-core decode replans where the native decode wrapper is actually supported.
- Guard zero-length recurrent checkpoint handling in `generator/job.py`.

### Architecture / model support
- Added / completed flashinfer-era support for:
  - `DeepseekV2ForCausalLM` / MLA-family aliases
  - `Qwen3_5ForConditionalGeneration`
  - `Qwen3_5MoeForConditionalGeneration`
  - `DeepseekVLV2ForCausalLM` (`deepseek-vl2` family)
- Added DeepSeek-VL2 EXL3 compile hooks needed to finish quantization.

### Multimodal fixes
- Stabilized the Qwen multimodal path (`Qwen3-VL`, `Qwen3.5`) after the backend transition.
- Fixed DeepSeek-VL2 image preprocessing / normalization so text+vision outputs are semantically correct.

### Benchmark tooling
- Updated `eval/perf.py` to use the current generator/job path instead of the older direct-forward assumptions.

### Functional validation
Confirmed working in local serving / smoke tests:
- `EXAONE-4.0.1` (text)
- `Qwen3-VL` (text + vision)
- `Qwen3.5-35B-A3B` (text + vision)
- `DeepSeek-V2-Lite-Chat` BF16 (text)
- `deepseek-vl2-tiny` BF16 and EXL3 4.0bpw (text + vision, single image, multi-image, grounding tags)
- `Qwen3-Next-80B-A3B` (text, after generator/parser fixes)
- `GLM-4.6V` (text + vision; see follow-up notes for remaining decode-path limits)

### Flash-attn vs flashinfer comparison (measured)
Representative tested points using the legacy flash-attn tree vs the current flashinfer tree:

#### Qwen3-Next-80B-A3B-Instruct-exl3-2.0bpw
Current flashinfer tree:
- Prefill: `256 = 196.19 tok/s`, `512 = 338.12 tok/s`
- Generate: `0 = 49.46 tok/s`, `256 = 49.29 tok/s`, `512 = 49.16 tok/s`

Legacy flash-attn tree:
- Prefill: `256 = 188.23 tok/s`, `512 = 323.89 tok/s`
- Generate: `0 = 40.32 tok/s`, `256 = 43.65 tok/s`, `512 = 43.26 tok/s`

At these measured points, the current flashinfer path is faster than the legacy flash-attn path.

#### GLM-4.6V-exl3-2.0bpw
Current flashinfer tree:
- Short-spot prefill: about `867 / 860 tok/s` (`256 / 512`)
- Generate: about `36 tok/s`

Legacy flash-attn tree:
- Prefill was comparable at short lengths, but the legacy generate benchmark path crashed later in the run.

So for GLM the migration is functionally stable, and the current tree now completes the tested generate path; deeper backend-specific tuning remains a follow-up.

#### deepseek-vl2-tiny-EXL3-4.0bpw
- Prefill:
  - `256 = 4332.33 tok/s`
  - `512 = 8488.48 tok/s`
  - `1024 = 8323.03 tok/s`
- Generate:
  - `0 = 112.37 tok/s`
  - `256 = 111.55 tok/s`
  - `512 = 111.71 tok/s`
  - `1024 = 111.53 tok/s`
- Real text/vision chat was also measured at roughly `105-111 tok/s`.

#### Qwen3.5-35B-A3B-EXL3-4.00bpw
- Prefill (spot check): about `240.61 tok/s`
- Generate (spot checks): about `35.31 / 35.12 tok/s`

#### Qwen3-VL-8B-Instruct-EXL3-4.0bpw
- Prefill (spot checks): about `3491.32 / 3430.09 tok/s`
- Generate (spot checks): about `57.52 / 57.38 tok/s`

## Important follow-up (not blocked by this PR)
The remaining major performance ceiling is now backend-specific, not migration-specific.

Two concrete examples:
- `EXAONE`: native FlashInfer non-tensor-core decode hits `Unsupported group_size: 5`
- `GLM-4.6V`: native FlashInfer non-tensor-core decode hits `Unsupported group_size: 12`

This means:
- the flashinfer migration is complete,
- but the last major gains for those models now require either:
  - backend-level FlashInfer support expansion, or
  - a separate high-risk decode-strategy project.

That work is intentionally left as a follow-up so this PR stays reviewable and keeps the hot path maintainable.

## Reviewer Focus
Please focus review on:
- flashinfer runtime path correctness and fallback invariants
- common-path performance changes (`attn.py`, `generator.py`, `job.py`)
- DeepSeek-VL2 architecture / quantization support
- Qwen multimodal stability after the backend transition
- any regressions in long-context behavior for supported fast paths

### DIFF
--- a/exllamav3/architecture/qwen3_5.py
+++ b/exllamav3/architecture/qwen3_5.py
@@ -16,15 +16,10 @@ from ..modules import (
     Linear,
     GatedDeltaNet,
     GatedMLP,
-    DeepstackEmbed,
 )
 from ..modules.attn import prepare_for_attn
 from ..modules.gated_delta_net import prepare_for_recurrence
 from .qwen3_vl import read_qwen3_vl_vision_config, read_qwen3_vl_pp_config, Qwen3VLVisionModel
-
-
-def conditional(condition, a, b):
-    return a if condition else b
 
 
 def read_qwen3_5_layer_types(config: Config, text_config_path: str, num_layers: int, full_attention_interval: int) -> list[str]:
@@ -111,8 +106,6 @@ class Qwen3_5Config(Config):
 
         self.vision_start_token_id = self.read_cfg(int, "vision_start_token_id", 151652)
         self.vision_end_token_id = self.read_cfg(int, "vision_end_token_id", 151653)
-        self.image_token_id = self.read_cfg(int, "image_token_id", None)
-        self.video_token_id = self.read_cfg(int, "video_token_id", None)
 
 
 class Qwen3_5MoeConfig(Config):
@@ -187,8 +180,6 @@ class Qwen3_5MoeConfig(Config):
 
         self.vision_start_token_id = self.read_cfg(int, "vision_start_token_id", 151652)
         self.vision_end_token_id = self.read_cfg(int, "vision_end_token_id", 151653)
-        self.image_token_id = self.read_cfg(int, "image_token_id", None)
-        self.video_token_id = self.read_cfg(int, "video_token_id", None)
 
 
 class Qwen3_5BaseModel(Model):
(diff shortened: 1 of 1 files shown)


lesj0610:
Qwen3.5-35B-A3B EXL3 quantization completed, then retested (English + Korean).

## What was fixed before retest
- `eval/quality_smoke_multilingual.py`
- `eval/quality_regression_en_zh.py`
- Applied HF chat template and propagated recurrent states through decode/NLL loops.

## Retest results (Qwen3.5-35B-A3B-EXL3-4.00bpw)
Model: `/ssd512g/models/Qwen3.5-35B-A3B-EXL3-4.00bpw`
GPU split: `23,23`

- decode-only speed (manual greedy path):
  - decode wrapper on: ~33.6-34.1 tok/s
  - decode wrapper off: ~28.9-29.3 tok/s
- generation quality:
  - English output is consistently gibberish / mixed random fragments
  - Korean output fails language constraint and is also gibberish (missing Hangul in smoke checks)

Examples from current outputs show repeated fragments such as `-sup`, mixed CJK/latin tokens, and no coherent answering behavior.

## Control check
DeepSeek-V2-Lite EXL3 under the same exllamav3 runtime still produces partially coherent outputs, so this is not just a generic test harness failure.

## Interim conclusion
Current Qwen3.5 path is integrated and runnable, but **E2E output quality is currently broken** in this branch state despite acceptable decode throughput. This indicates a likely architecture/tensor-mapping/runtime correctness issue for Qwen3.5 rather than a pure performance problem.

I will continue with root-cause narrowing in follow-up commits.

turboderp:
(Thanks, by the way, this is great work.) (:

lesj0610:
I agree that forcing FlashInfer as the only/default backend is too aggressive for upstream right now.

A better direction would be to keep FlashAttention and FlashInfer side-by-side, and select the best supported backend at initialization based on:
- which dependencies are actually installed,
- the model’s capabilities and constraints,
- and the stability/performance characteristics of each path.

In practice, that means:
- only expose backend choices that are actually available in the environment,
- keep backend selection fixed at model/layer init time (not per-token),
- use FlashInfer where it is required or clearly beneficial (for example MLA-capable models),
- keep the existing stable paths where FlashInfer still has coverage gaps,
- and retain SDPA as the universal fallback.

So I do not think upstream should be “FlashInfer-only” today, but I do think it should move toward a capability-based backend abstraction where FlashInfer and FlashAttention can coexist cleanly.

PR #159, opened 2026-03-02
feat(qwen): add Qwen3.5 architecture support

lesj0610:
Summary:
- add Qwen3.5 architecture support
- reuse the Qwen3-Next path where appropriate
- include the minimal Qwen multimodal/runtime changes required for correctness

This is intentionally split out from the larger flashinfer migration work.


lesj0610:
Thanks — treating #159 as the new baseline makes sense. I will rebase the remaining Qwen3.5 changes on top of `dev` and split them into smaller follow-up PRs instead of re-posting the full support patch.

The remaining work is now limited to:
- a small Qwen3.5-specific follow-up for the remaining multimodal wiring differences, and
- a separate linear-attention follow-up for the remaining `gated_delta_net` / `gated_rmsnorm` stability changes.

That should make the remaining review scope much smaller and keep `dev` as the source of truth for the base Qwen3.5 implementation.

PR #163, opened 2026-03-05
generator: always reconstruct held text when replacement chars appear

lesj0610:
## Summary
- Hold text whenever `�` appears and always attempt reconstruction from held token IDs.
- Removes the previous `1 <= count(�) < 5` threshold that could leak broken streamed text for some byte-fallback boundaries.
- Keeps behavior scoped to generator text assembly only (no convert/quant changes).

## Why
Some outputs could temporarily contain exactly five replacement characters in `held_text`, which skipped reconstruction and emitted broken text (e.g. `�����`) even though decoding the same token IDs was valid.

## Validation
- Reproduced stream/decode mismatch before patch and confirmed match after patch in local regression checks.
- Built and smoke-tested runtime paths used during verification without touching conversion/quantization code.

### DIFF
--- a/exllamav3/generator/job.py
+++ b/exllamav3/generator/job.py
@@ -652,8 +652,9 @@ class Job:
         if filter_eos_condition:
             return emit(results, emit_eos = True, emit_held = True, eos_reason = "end_filter")
 
-        # Hold text if it contains an incomplete character
-        if 1 <= self.held_text.count("�") < 5:
+        # Hold text if it contains an incomplete character. Always attempt
+        # reconstruction when replacement characters are present.
+        if "�" in self.held_text:
             test_decode = self.generator.tokenizer.decode(
                 self.held_tokens.torch(),
                 decode_special_tokens = self.decode_special_tokens


turboderp:
Not sure I follow the logic here. You should never get more than four replacement characters unless sampling has broken. A BPE model produces at least one byte per token, and a UTF8 character is at most 4 bytes. So if you end up with 5 replacement characters (5 invalid bytes) due to broken inference or sampling, you could never recover from that by appending more tokens to the held string.

lesj0610:
You're right if the count referred to raw invalid UTF-8 bytes. In this path it doesn't: `held_text` is the fast-path string assembled from `id_to_piece[next_token]`, and only afterwards compared against `tokenizer.decode(self.held_tokens)`.

So the previous `1 <= count("�") < 5` check was using the replacement-character count in that intermediate piecewise string as a proxy for "recoverable by full decode". That proxy turned out to be too strict.

In the failing cases I saw, the piecewise `held_text` contained 5 replacement chars and would be emitted as broken text, but decoding the exact same accumulated token IDs via `tokenizer.decode(self.held_tokens)` produced valid output. The user-visible regression was Korean/CJK streaming text leaking as `�����...`, not a permanently broken sample that stayed invalid under full decode.

So this change is not trying to recover impossible broken sampling. It only says: if the fast path contains replacement chars at all, verify against the authoritative full decode before emitting. If the full decode still contains replacement chars, behavior is unchanged apart from the existing `> 20` guard to avoid holding forever.

In other words, the 4-byte UTF-8 argument applies to raw byte sequences, but `held_text.count("�")` here is counting replacement chars in the incremental piece assembly, which is a different thing.

If it helps, I can also add a focused regression test around the piecewise-vs-full-decode mismatch instead of relying on the Korean repro alone.

PR #185, opened 2026-04-03
feat(gemma4): add Gemma4 multimodal architecture support

lesj0610:
## Summary
- add cumulative `Gemma4ForConditionalGeneration` support for Gemma4 text and vision components
- split Gemma4 multimodal and vision-specific logic into the dedicated adapter file `exllamav3/architecture/gemma4_mm.py`
- keep Gemma4 text architecture wiring in `exllamav3/architecture/gemma4.py` and keep vision preprocessing, patch conversion, vision embedding extraction, and vision-group setup isolated in the MM adapter layer
- add Gemma4 multimodal attention handling needed for image-token groups in sliding-attention layers
- keep multimodal prefill atomic across image embedding spans
- allow `hf_chat_template(..., embeddings=[...])` to replace rendered image placeholders with MM embedding aliases automatically
- extend Gemma4 smoke coverage for vision embedding extraction and multimodal generation
- intentionally keep turboquant / reduced-SWA / cache-path work out of this PR so the base support review stays focused

## Why The MM Split Exists
The Gemma4 multimodal path was intentionally moved into a separate file, `exllamav3/architecture/gemma4_mm.py`, instead of leaving it embedded in the main Gemma4 text architecture file.

This split is intentional for reviewability and long-term maintenance:
- it keeps the text architecture adapter focused on Gemma4 text-model wiring and base model capabilities
- it keeps vision preprocessing, image patch conversion, pooled vision embedding generation, and multimodal token-group bookkeeping in one isolated MM-specific module
- it reduces the chance that later cache/runtime work gets mixed into the core model-support PR
- it makes the Gemma4 MM support easier to review independently from the follow-up turboquant/cache optimization work

In short, the split is there to keep the base Gemma4 model-support PR narrow, readable, and maintainable, and the dedicated `gemma4_mm.py` file is part of this PR's actual patch set.

## Validation
- `python tests/smoke_gemma4_arch.py --model_dir <Gemma4 31B checkpoint> --device cuda:0`
- `python tests/smoke_gemma4_arch.py --model_dir <Gemma4 31B quantized output> --device cuda:0`
- `python tests/smoke_gemma4_arch.py --model_dir <Gemma4 31B quantized output> --device cuda:0 --multimodal_generation --reserve_per_device 0`
- `python tests/smoke_gemma4_mm.py --model_dir <Gemma4 31B quantized output> --reserve_per_device 0`
- `python tests/smoke_gemma4_arch.py --model_dir <Gemma4 26B-A4B checkpoint> --device cuda:0`
- `python tests/smoke_gemma4_arch.py --model_dir <Gemma4 26B-A4B quantized output> --device cuda:0`
- `python tests/smoke_gemma4_arch.py --model_dir <Gemma4 26B-A4B quantized output> --device cuda:0 --full_load --reserve_per_device 0.5,0.5`
- `python tests/smoke_gemma4_mm.py --model_dir <Gemma4 26B-A4B quantized output> --reserve_per_device 0`
- support-only 31B sequential text generation sanity check on the current PR head
- EXL3 conversion completed for the Gemma4 31B checkpoint and the Gemma4 26B-A4B checkpoint
- Gemma4 26B-A4B EXL3 conversion was validated with parallel MoE quantization mode (`-pm`)

## Scope
- validated Gemma4 text generation on 31B dense and 26B-A4B MoE checkpoints and EXL3 outputs
- validated Gemma4 multimodal generation on Gemma4 31B IT and Gemma4 26B-A4B IT
- Gemma4 vision inputs are supported through exllamav3 MM embeddings
- this PR is the Gemma4 model-support and multimodal-support layer
- turboquant / reduced-SWA / cache-path work is intentionally being kept in a separate follow-up PR

### DIFF
--- a/exllamav3/architecture/architectures.py
+++ b/exllamav3/architecture/architectures.py
@@ -9,6 +9,7 @@ from .ernie4_5_moe import Ernie4_5MoEModel
 from .exaone4 import Exaone4Model
 from .gemma2 import Gemma2Model
 from .gemma3 import Gemma3Model, Gemma3TextModel
+from .gemma4 import Gemma4TextModel
 from .glm4 import Glm4Model
 from .glm4_moe import Glm4MoeModel
 from .glm4v import Glm4VModel
@@ -58,6 +59,7 @@ ARCHITECTURES = {
         Gemma2Model,
         Gemma3Model,
         Gemma3TextModel,
+        Gemma4TextModel,
         Glm4Model,
         Glm4MoeModel,
         Glm4VModel,
--- a/exllamav3/architecture/gemma4.py
+++ b/exllamav3/architecture/gemma4.py
@@ -1,13 +1,10 @@
 from __future__ import annotations
 
 import json
-import math
 import os
 from types import SimpleNamespace
 
-import numpy as np
 import torch
-from PIL import Image
 from typing_extensions import override
 
 from ..model.config import Config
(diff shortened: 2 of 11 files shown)


Ph0rk0z:
I ran the unquantized transformers model with a chat template from nvidia and I think I can get a single message off before it starts to reply with gibberish. That's the 31b so not MoE.

lesj0610:
@Ph0rk0z I rechecked the `31B` concern on the current branch from the exllamav3 side.

I was not able to reproduce a `31B` "single message, then gibberish" failure with the current implementation. On `gemma-4-31B-it-exl3-4bpw`, reusing the same `Generator`/`Cache` across repeated requests, I get stable sequential outputs for both text and MM prompts:

- `Hi.`
- `alpha alpha alpha`
- `Curious.`
- `Hi.`
- `alpha alpha alpha`
- `Curious.`
- `Cat`
- `Strawberry`
- `first`
- `yes`
- final: `{"status":"OK","nreq":10}`

I also have a longer `31B` mixed-soak passing at `nreq=256` on the current branch.

So the earlier `31B` gibberish report does not currently reproduce for me in exllamav3. Since your comment mentions an unquantized Transformers + NVIDIA chat template run, that may be a different failure mode than the current exllamav3 path itself.

Ph0rk0z:
I simply didn't quantize the model but exl3 allows to load the full transformer model for testing and it should work correctly as it does for others.

lesj0610:
Thanks, that's helpful context. Since there have been several follow-up fixes after the initial Gemma4 support work, can you confirm whether you tested the latest cumulative PR state, or an earlier snapshot of the implementation?

If you still reproduce it on the latest state, could you share the exact commit, model artifact, load args, and a minimal prompt sequence? That would help separate a current support-path issue from an already-fixed earlier state.

Ph0rk0z:
I tested as of https://github.com/turboderp-org/exllamav3/pull/185/changes/94981746a9960e9a4f207350fe24e1d9fe58956f

Can test again and maybe do full cache since your new PR implies that cache quantization was not accounted for in the PR.

turboderp:
There's probably not much point in testing it. I've already merged the PR (or some older version) and spent several days completely reworking the implementation. It'll all be pushed to `dev` very soon, and so much has changed that you won't really be testing the same thing.

(Overall, this is an incredibly invasive spaghetti architecture, and I would normally just give it a pass except everyone is super hyped for it, so I guess here we are.)

Ph0rk0z:
When I used the model on API it was pretty nice. But so far most implementations have been broken in one way or another. Hopefully the honeymoon doesn't end after more time with it. The logprobs are pretty slanted to favor only one token.

turboderp:
Is in dev branch now. Might need some more work, and definitely needs a custom attention kernel to maintain speed on longer contexts. But I've got something coming up. In the meantime here are quants: https://huggingface.co/turboderp/gemma-4-26B-A4B-exl3

31B is converting now, and I'll be back on this tomorrow.

lesj0610:
At this point I think this PR is effectively superseded by the Gemma4 support/MM work that has already landed on `dev` in upstream form.

In other words, the current situation no longer seems to be "this PR adds Gemma4 support that `dev` does not have". The remaining delta here is mostly the later MM file split / organization work, not the original support path itself.

So instead of trying to keep this PR alive through the current conflicts, I plan to follow up with a much smaller PR against the current `dev` branch that only carries the remaining MM split/cleanup work on top of the maintainer's current Gemma4 implementation.

Because of that, I think closing this PR is the right next step.

turboderp:
Maybe I should clarify that this *is* what landed in `dev`. I'm not great with github, but I think that once `dev` gets merged into `main` (need to validate a few more things) this will register as merged.

PR #188, opened 2026-04-08
Gemma4: add turboquant KV cache with reduced SWA and soak validation

lesj0610:
> Note: because upstream PR #185 is still based on the fork branch `lesj0610:feat/gemma4-support`, GitHub cannot show a truly stacked upstream PR on top of that branch.

> This upstream draft therefore targets `dev` for visibility, but the intended logical review target is the turboquant delta on top of PR #185. A clean stacked draft against `feat/gemma4-support` also exists in the fork: https://github.com/lesj0610/exllamav3/pull/5

# PR Draft: Gemma4 turboquant KV cache with reduced SWA and Gemma4-local soak fixes

## Summary

This PR adds a Gemma4-specific turboquant KV-cache path with reduced SWA
capacity, role-aware page-table handling, and Gemma4-local stability fixes for
mixed text/MM serving.

This is intentionally **not** the cumulative Gemma4 support PR. `PR #185`
should contain the squashed first-stage Gemma4 support work:

- initial Gemma4 architecture support
- multimodal / vision support
- support-side Gemma4/MM fixes

This turboquant PR is the follow-up stacked on top of that support PR. It only
covers the Gemma4 cache, page-table, and runtime changes that are needed to
make the turboquant path work correctly and efficiently once Gemma4 support
exists.

## Reference snapshots used for comparison

The implementation comparisons in this PR were made against the following local
repository snapshots:

- `exllamav3`: `84004e6` (`v0.0.28-4-g84004e6`)
- `vLLM`: `1a2c17634` (`v0.19.1rc0-102-g1a2c17634`)
- `llama.cpp`: `d9a12c82f` (`b8708-9-gd9a12c82f`)

These are local working snapshots rather than release tags, so the comparison
should be read as “what was implemented in those codebases at the time this PR
was prepared,” not as a claim about all earlier or later revisions.

## Relation to other runtimes

Gemma4 already requires architecture-specific handling in other inference
engines.

- **vLLM** does not keep Gemma4 on a generic path. It has a dedicated Gemma4
  implementation and, when `kv_sharing_fast_prefill` is enabled, splits
  prefill into `Gemma4SelfDecoderLayers` and `Gemma4CrossDecoderLayers` around
  `num_kv_shared_layers`. The main optimization there is fast prefill rather
  than a reduced local KV-cache design.

- **llama.cpp** also treats Gemma4 as a dedicated ISWA model. It maintains
  separate base and SWA KV caches, builds separate `k_idxs / v_idxs / kq_mask`
  inputs for those caches, and sizes the SWA cache independently from the full
  cache. Conceptually, this is the closest external design to the one used
  here.

This PR follows the same overall direction: Gemma4 is handled through
architecture-specific hooks instead of forcing it through the generic KV-cache
assumptions. In exllamav3, that takes the form of:

- Gemma4-specific quantized KV cache layers
- Gemma4-specific reduced-SWA sizing
- Gemma4-specific role-aware page-table handling for `full` and `swa`
- Gemma4-specific compact KV fast paths
- Gemma4-local mixed text/MM stability fixes

So this PR is not introducing a special case in isolation; it brings exllamav3
in line with the fact that Gemma4 already needs dedicated cache/runtime
treatment in other mature runtimes as well.

## Why this should live upstream

Gemma4 does not fit the generic EXL3 KV-cache assumptions cleanly:

- Gemma4 mixes full/global attention with sliding-window attention.
- The sliding-window view needs a smaller local cache budget than the full
  cache budget.
- Mixed text/MM traffic needs role-aware cache-copy handling between the full
  and SWA views.
- The generic one-cache page model is not enough to preserve correctness for
  Gemma4 prompt reuse and local-window restores.

Without these changes, the Gemma4 turboquant path either wastes a large amount
of KV memory or becomes unstable under mixed text/MM churn.

## Scope

This PR adds:

- Gemma4-specific quantized KV cache layers
- Gemma4-specific reduced-SWA sizing
- Gemma4-specific role-aware page-table handling for `full` and `swa`
- Gemma4-specific compact KV fast paths
- Gemma4-specific stability fixes for mixed text/MM serving

This PR does **not** add:

- generic EXL3 serving bugfixes for unrelated models
- Gemma4 MM architecture split or vision-model support
- TP enablement for `26B-A4B`

## Non-Gemma impact

The shared changes that remain in this PR are limited to generic extension
hooks that stay on the original default path for non-Gemma models.

### What non-Gemma models do **not** do

Non-Gemma models do not opt into any of the Gemma4-specific runtime classes:

- `model.caps["page_table_cls"]` is unset
- `model.caps["cache_layer_selector"]` is unset
- `model.caps["quantized_kv_cache_layer"]` is unset

So non-Gemma models continue to use:

- the default `PageTable`
- the default `CacheLayer_quant`
- the original single-cache block-table path

### What was actually checked

I re-ran short quantized-cache generation probes on two non-Gemma models using
the default `Cache`, `Generator`, and `Job` APIs:

- `EXAONE-4.0.1-32B-EXL3-4.00bpw`
- `Magistral-Small-2509-EXL3-4.0bpw`

Observed results after the latest shared-surface cleanup:

- `page_table_cls = null`
- `cache_layer_selector = false`
- `quantized_kv_cache_layer = null`
- instantiated cache layers were plain `CacheLayer_quant`
- both models completed short quantized-cache generation successfully

Example probe outputs:

```json
{"model": "exaone", "page_table_cls": null, "cache_layer_selector": false, "quantized_kv_cache_layer": null, "cache_layer_types": ["CacheLayer_quant"], "output": "\\n\\nAnswer: Yes"}
{"model": "magistral", "page_table_cls": null, "cache_layer_selector": false, "quantized_kv_cache_layer": null, "cache_layer_types": ["CacheLayer_quant"], "output": " What is the name of"}
```

The exact text is not important here; the point of the probe is that both
non-Gemma models stay on the default quantized-cache path and complete actual
generation without opting into any Gemma4-specific classes.

I also re-ran the shared `model_init.init(...)` path on a non-Gemma model with
`cache_quant="4"` and verified that it still resolves to:

```json
{"page_table_cls": null, "cache_layer_selector": false, "quantized_kv_cache_layer": null, "cache_layer_types": ["CacheLayer_quant"]}
```

So the remaining `model_init` change is still a default-preserving constructor
hook rather than a non-Gemma runtime behavior change.

### Why the remaining shared hooks are acceptable

The remaining shared changes are:

- constructor-time cache-layer selection in `Cache`
- constructor-time page-table selection in `Generator`
- generic `PageTable` extension points used by custom pagetables

For non-Gemma models these hooks stay on the default implementation and do not
perform any Gemma4-specific tensor work, SWA bookkeeping, or extra cache
allocations.

In particular:

- `Generator` now chooses the pagetable class once at construction time.
  Non-Gemma models still instantiate the default `PageTable`.
- `Job` only talks to the generic `PageTable` interface and no longer branches
  on Gemma4-specific runtime types.
- `Cache.copy_page(..., page_plan=...)` keeps the original path unless a custom
  pagetable explicitly passes a role-aware copy plan. Non-Gemma models never do.

To keep the shared surface small, this PR intentionally does **not** add a
global `model_init --swa_cache_size` knob. Reduced-SWA configuration stays on
the Gemma4 cache path itself.

## Turboquant vs support-only baseline

To make the tradeoff concrete, I added a dedicated Gemma4 comparison harness:

- `tests/bench_gemma4_compare.py`

This compares the current turboquant branch against the support-only Gemma4
branch in isolated subprocesses so both trees can be measured with the same
workload.

The comparison below was run on the `31B` EXL3 model with:

```bash
python tests/bench_gemma4_compare.py \
  --baseline_repo_root /tmp/exllamav3_support_split \
  --current_repo_root . \
  --model_dir /ssd512g/models/gemma-4-31B-it-exl3-4bpw \
  --cache_bits 4 \
  --cache_size 4096 \
  --swa_cache_size 1024 \
  --text_repeats 2 \
  --mm_repeats 2
```

Observed result:

```json
{
  "baseline": {
    "cache_footprint_mib": 3520.0,
    "text_repeat_times_s": [1.518, 0.1036],
    "dup_same_times_s": [6.0431, 0.5146],
    "pair_animal_s": 3.2601
  },
  "current": {
    "cache_footprint_mib": 315.86,
    "text_repeat_times_s": [0.9159, 0.8855],
    "dup_same_times_s": [0.8121, 0.7562],
    "pair_animal_s": 0.7508
  },
  "comparison": {
    "cache_ratio_baseline_over_current": 11.1442,
    "cache_reduction_pct": 91.03,
    "text_first_speedup": 1.6574,
    "text_second_speedup": 0.117,
    "dup_same_first_speedup": 7.4413,
    "dup_same_second_speedup": 0.6805,
    "pair_animal_speedup": 4.3422
  }
}
```

The main takeaway is:

- turboquant reduces Gemma4 KV-cache footprint by about `91%` in this setup
- turboquant materially improves cold MM latency in the mixed-image workloads
  above
- the support-only full-cache baseline can still win on fully warmed text/MM
  reuse, which is expected because it avoids quantized-KV overhead and keeps a
  much larger cache resident

So the value of this PR is not “turboquant beats the full-cache path at every
microbenchmark.” The value is that Gemma4 gets a much smaller cache footprint,
better first-run MM behavior, and stable mixed text/MM serving without pushing
those costs onto non-Gemma models.

### Regression tests

```bash
python -m pytest -q \
  tests/test_gemma4_turboquant.py \
  tests/test_gemma4_kv_ext.py
```

Current result:

- `63 passed, 14 warnings`
- all warnings came from the same upstream Torch deprecation:
  `torch.jit.script_method`

### Latest follow-up on the current PR head

After the initial draft was opened, the current head also addressed two
speculative generator correctness issues raised during review:

- rejected speculative tokens now roll back KV positions per sequence again
- Gemma4 draft decode now advances `cache_seqlens_full` and
  `cache_seqlens_swa` together with the default `cache_seqlens`

### Gemma4 focused repros

`26B-A4B` previously had a remaining mixed text/MM corruption where a fresh
plain-text prefill could collapse to `<pad>...` after earlier MM churn.

That path is now covered by focused runtime repros:

- fresh text forced onto later full-cache pages now returns `Inquisitive`
- `pair -> drop snapshots -> fruit -> text` now returns:
  - `pair -> First`
  - `fruit -> Strawberry`
  - `text -> Inquisitive`

### Gemma4 smoke / soak

- `31B` MM smoke: pass
- `31B` mixed soak, `seed=20260407`, `nreq=256`: `{"status":"OK","nreq":256}`
- `26B` mixed soak, `seed=20260407`, `nreq=256`: `{"status":"OK","nreq":256}`
- `26B` mixed soak, `seed=20260407`, `nreq=512`: `{"status":"OK","nreq":512}`
- earlier hard soak coverage also includes:
  - `swa=768`
  - `swa=1024`
  - dynamic prefix-churn / text-MM mode flips
  - `max_num_tokens = 262144`

## Notes on the final 26B fix

The final remaining `26B` corruption bug was fixed inside
`Gemma4PageTable.allocate_pages()`.

The bug turned out to be Gemma4-local: a fresh full-prefill could break if it
started on later full-cache page indices after prior churn. The fix keeps fresh
Gemma4 full-prefill on the stable low-index full pages, without changing the
non-Gemma page-table path.

## Known limitations

- `26B-A4B` multimodal duplicate-reasoning quality is still a weight-side model
  issue. That is separate from the fixed runtime corruption bug.
- `26B-A4B` TP semantics are still not ready to expose.

### DIFF
--- a/exllamav3/architecture/architectures.py
+++ b/exllamav3/architecture/architectures.py
@@ -9,6 +9,7 @@ from .ernie4_5_moe import Ernie4_5MoEModel
 from .exaone4 import Exaone4Model
 from .gemma2 import Gemma2Model
 from .gemma3 import Gemma3Model, Gemma3TextModel
+from .gemma4 import Gemma4TextModel
 from .glm4 import Glm4Model
 from .glm4_moe import Glm4MoeModel
 from .glm4v import Glm4VModel
@@ -58,6 +59,7 @@ ARCHITECTURES = {
         Gemma2Model,
         Gemma3Model,
         Gemma3TextModel,
+        Gemma4TextModel,
         Glm4Model,
         Glm4MoeModel,
         Glm4VModel,
--- a/exllamav3/architecture/gemma4.py
+++ b/exllamav3/architecture/gemma4.py
@@ -1,13 +1,10 @@
 from __future__ import annotations
 
 import json
-import math
 import os
 from types import SimpleNamespace
 
-import numpy as np
 import torch
-from PIL import Image
 from typing_extensions import override
 
 from ..model.config import Config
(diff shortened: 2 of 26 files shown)


Ph0rk0z:
exllama already does hadamard transforms on KV cash afaik. the only thing that had to be done is making it work with the funky gemma layer/cache structure.

lesj0610:
@Ph0rk0z Thanks. The issue here was not whether exllama already has generic Hadamard / KV primitives.

The harder part was making Gemma4's full/SWA cache structure behave correctly under mixed text/MM serving while keeping the non-Gemma default path as unchanged as possible. A large part of the work here was specifically about reducing shared-surface impact, not expanding it.

That is also why the cumulative Gemma4 support work and the turboquant follow-up were split, and why the remaining shared hooks were kept to default-preserving constructor/interface points rather than Gemma4-specific runtime behavior in non-Gemma paths.

If there is a smaller concrete implementation that preserves those same constraints and still passes the same mixed-soak and correctness cases, I'd be happy to look at it.

PR #189, opened 2026-04-09
LoRA: Add PEFT adapter support

mayakwd:
Adds runtime LoRA (Low-Rank Adaptation) support for PEFT-format adapters. LoRA weights are applied on top of dequantized output of each target linear layer without modifying base model weights.

Multiple adapters can be loaded simultaneously — the deltas are summed during forward pass. Each adapter can be independently unloaded to revert its contribution.

**Key implementation details:**

PEFT keys are matched against model modules by suffix rather than hardcoding prefix patterns, so any PEFT key format (standard, Qwen3.5/multimodal, etc.) works automatically. LoRA A/B matrices are pre-transposed and pre-scaled at load time to avoid per-token overhead. Quantized layers that pad feature dimensions to block-size multiples get their LoRA matrices zero-padded to match. Tensor-parallel sliced modules are skipped with a warning since they require per-slice LoRA handling.

**What I tested:**

Tested with Qwen3.5-27B-exl3-4.0bpw and an Unsloth-trained PEFT adapter (r=32, alpha=64, 512 tensors). Verified that generation output differs with LoRA active vs base model, and that unloading the adapter restores base model behavior.

### DIFF
--- a/exllamav3/model/lora.py
+++ b/exllamav3/model/lora.py
@@ -1,0 +1,221 @@
+"""
+LoRA (Low-Rank Adaptation) support for ExLlamaV3.
+
+Loads PEFT-format LoRA adapters and applies them at runtime without
+modifying base model weights. Multiple adapters can be loaded
+simultaneously; all loaded adapters are applied during forward pass.
+
+Usage::
+
+    lora = LoRA.from_directory(model, "/path/to/peft-adapter")
+    # All generation now includes this adapter's contribution
+    response = generator.generate(prompt = "Hello", ...)
+    # Unload to revert to base model
+    lora.unload()
+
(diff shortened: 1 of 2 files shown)


turboderp:
Thanks. :+1: Sorry for being slow. (Merged into dev, merging dev into main soon.)

mayakwd:
Thank you, @turboderp, and thanks for such a great inference library!

I have also drafted MTP-head (Multi-Token Prediction) support for Qwen. It gives +10-20% tok/s, and some other improvements/optimizations. 

I'm not sure those are reasonable contributions, since you are the only maintainer of exllamav3 and might have other priorities. I'll be happy to learn a bit about what's on your mind regarding ongoing priorities.

Cheers.

PR #194, opened 2026-04-16
Detailed size estimation + fix #193 (overestimating fused QKV storage)

### LINKED ISSUE #193: [Bug] storage_size (TP) overestimates allocation of fused linear layers like QKV / QKVZ
Looking at https://github.com/turboderp-org/exllamav3/blob/cb1a436f8f948339e4468c48e4e4e9640c2241da/exllamav3/modules/linear.py#L430-L463

For fused layers `self.key` is none and `self.is_exl3_storage(self.key):` branch will be skipped in favor of the Fp16 computation `2 * self.in_features * self.out_features`

However this is significantly overestimating as the true storage cost is `bpw/8 * self.in_features * self.out_features`. For 3bpw that would be a 5.33x overestimation.

This concerns:
- vision tower of Qwen3.5, Qwen3-VL, Qwen2.5-VL, GLM-4.xV
- Phi3
- Qwen3.5 and Qwen-Next Gated Delta Net (that don't support tensor parallelism)

As far as I've seen it is only used for tensor parallelism allocation computation which is disabled on Qwen-Next / Qwen3.5 anyway

mratsim:
This PR does 2 things related to size:
1. Fix #193, for fused QKV layers, when `key_q`, `key_k`, `key_v` are not set, self.key is undefined and storage will used the FP16 formula here: https://github.com/turboderp-org/exllamav3/blob/cb1a436f8f948339e4468c48e4e4e9640c2241da/exllamav3/modules/linear.py#L430-L463

    This can be fixed simply by setting those keys at the model architecture level.

2. This expands `model.py` with `get_detailed_weights_info` that is then used by `size_estimation.py` for detailed statistics.

I noticed 2 things:
- Convolution has no weights_numel set if weights are not loaded so I added a formula to compute it from metadata. 
- Linear doesn't count the number of elements in the bias, that might affect TP allocation and might be a bug. I don't count the number of elements in convolution bias to mirror Linear layers.

### DIFF
--- a/exllamav3/architecture/glm4v.py
+++ b/exllamav3/architecture/glm4v.py
@@ -212,7 +212,10 @@ class Glm4VVisionModel(Model):
                             rope_style = RopeStyle.NEOX,
                         ),
                         key_fused_qkv = "qkv",
-                        key_o = "proj",
+                        key_q="q_proj",
+                        key_k="k_proj",
+                        key_v="v_proj",
+                        key_o = "o_proj",
                         qmap = "block.attn",
                     ),
                     mlp_norm = RMSNorm(
--- a/exllamav3/architecture/qwen2_5_vl.py
+++ b/exllamav3/architecture/qwen2_5_vl.py
@@ -224,8 +224,11 @@ class Qwen2_5VLVisionModel(Model):
                             head_dim = v.head_dim,
                             rope_style = RopeStyle.NEOX,
                         ),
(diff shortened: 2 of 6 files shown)


mratsim:
The output of size_estimation.py looks like this now
image

turboderp:
I think this has gone a little stale. I like the size estimation though, probably worth rebasing.

PR #206, opened 2026-05-08
MTP speculative decoding for Qwen3.5 / Qwen3.6

alexhunt7:
## Multi-Token Prediction (MTP) speculative decoding for Qwen3.5 / Qwen3.6

Adds integrated MTP draft model support for Qwen3.5/3.6 models, enabling speculative decoding via the native MTP head shipped alongside these models.

### Overview

Introduces `Qwen3_5MTPDraftModel`, a synthetic draft architecture that loads only the `mtp.*` tensors from the original BF16 checkpoint and borrows `embed_tokens` / `lm_head` from the target model via weakrefs. The MTP head is a single transformer block plus a 2H->H projection over `[embed(token), hidden]`, supporting recurrent multi-step drafting (NEXTN-style).

Unlike standard draft models, MTP requires the target's **pre-final-norm** hidden state (residual stream after the last transformer block, before `output_norm`). Feeding post-norm would double-normalize and hurt acceptance rates.

### Changes

**Architecture** (`qwen3_5_mtp.py`):
- `attach_to()` binds to target model, marks last `TransformerBlock` to export its pre-final-norm residual stream
- `step()` embeds token, RMSNorms both embedding and hidden, concatenates, projects through `fc`, runs MTP block, returns residual hidden + logits
- `update_kv_from_target()` backfills MTP K/V cache from target hidden states (slot p stores `(token@(p+1), hidden@p)`)

**Generator** (`generator.py`):
- `iterate_draftmodel_mtp_gen()` — recurrent drafting loop; one MTP step per draft token, writing K/V at correct cache slots (lagging target by 1)
- Post-verification: captures `export_states` to seed next round's `last_hidden`, backfills accepted-position K/V with target-actual values
- Fixes pre-existing `states[idx]` → `states[j]` indexing bug in recurrent state rewind (only affected models with `recurrent_cache`)

**Prefill** (`job.py`):
- After each prefill chunk, backfills MTP draft cache from exported hidden states
- Threads `seq.mtp_carry_hidden` across chunks for slot continuity

**Config** (`config.py`):
- `arch_override` parameter on `Config.from_directory()` allows loading a Qwen3.5/3.6 BF16 directory as `Qwen3_5MTPDraftModel`, picking up only `mtp.*` tensors despite `config.json` declaring `Qwen3_5ForConditionalGeneration`

**CLI** (`model_init.py`):
- New `-dm_arch` / `--draft_arch_override` flag

**Other**:
- Registers `mtp_num_hidden_layers` / `mtp_use_dedicated_embeddings` in Qwen3_5 config (informational; head loaded as separate draft model)
- Adds `mtp_carry_hidden` slot on `Sequence` for cross-chunk hidden threading
- `LinearEXL3`: removes unnecessary `rows <= 32` threshold check on fast path

### Limitations

- Multi-sequence jobs (CFG) not supported with MTP draft
- Multimodal embeddings not supported with MTP draft
- `update_kv_from_target()` currently supports a single MTP layer (Qwen3.6 default)
- MTP draft model loaded in BF16 (not yet supported by exl3 quant converter)

### AI Disclosure
Most of this was written by AI (a combination of Claude Opus and Qwen3.6 27B) under my guidance. I did steer them and read through it to ensure they weren't doing anything obviously stupid, but I don't have the math or ML background to verify their algorithm changes are correct. I can only verify that it works empirically.

I have tested this locally with tabbyAPI ([with some changes to pass the new arguments](https://github.com/theroyallab/tabbyAPI/pull/421)), and I get a more than 2x token generation speedup on my 5070ti + 3060ti. The output is coherent and indistinguishable to me from without MTP. Prompt processing does take a bit of a hit though (~33% slower).

### DIFF
--- a/exllamav3/architecture/architectures.py
+++ b/exllamav3/architecture/architectures.py
@@ -35,6 +35,7 @@ from .qwen2 import Qwen2Model
 from .qwen2_5_vl import Qwen2_5VLModel
 from .qwen3 import Qwen3Model
 from .qwen3_5 import Qwen3_5Model, Qwen3_5MoeModel, Qwen3_5VLModel, Qwen3_5VLMoeModel
+from .qwen3_5_mtp import Qwen3_5MTPDraftModel
 from .qwen3_moe import Qwen3MoeModel
 from .qwen3_next import Qwen3NextModel
 from .qwen3_vl import Qwen3VLModel
@@ -91,6 +92,7 @@ ARCHITECTURES = {
         Qwen3_5MoeModel,
         Qwen3_5VLModel,
         Qwen3_5VLMoeModel,
+        Qwen3_5MTPDraftModel,
         Qwen3MoeModel,
         Qwen3NextModel,
         Qwen3VLModel,
--- a/exllamav3/architecture/qwen3_5.py
+++ b/exllamav3/architecture/qwen3_5.py
@@ -96,6 +96,10 @@ class Qwen3_5VLBaseConfig(Config):
             self.full_attention_interval,
         )
 
+        # MTP (multi-token prediction) head — informational; head loaded as separate draft model
+        self.mtp_num_hidden_layers = self.read_cfg(int, pfx("mtp_num_hidden_layers"), 0)
+        self.mtp_use_dedicated_embeddings = self.read_cfg(bool, pfx("mtp_use_dedicated_embeddings"), False)
+
         # RoPE
         self.rope_settings = self.read_rope_settings_default(
             RopeStyle.NEOX,
@@ -212,6 +216,10 @@ class Qwen3_5VLMoeBaseConfig(Config):
             self.num_hidden_layers,
             self.full_attention_interval,
         )
+
+        # MTP (multi-token prediction) head — informational; head loaded as separate draft model
+        self.mtp_num_hidden_layers = self.read_cfg(int, pfx("mtp_num_hidden_layers"), 0)
+        self.mtp_use_dedicated_embeddings = self.read_cfg(bool, pfx("mtp_use_dedicated_embeddings"), False)
 
         # RoPE
         self.rope_settings = self.read_rope_settings_default(
--- a/exllamav3/architecture/qwen3_5_mtp.py
+++ b/exllamav3/architecture/qwen3_5_mtp.py
@@ -1,0 +1,472 @@
+from __future__ import annotations
+from typing_extensions import override
+import torch
+import weakref
+
+from ..cache import Cache
+from ..ext import exllamav3_ext as ext
+from ..model.config import Config, no_default
+from ..model.model import Model
+from ..util.rope import RopeStyle
+from ..modules import RMSNorm, Embedding, TransformerBlock, Attention, GatedMLP, Linear
+from ..modules.attn import prepare_for_attn
+from ..modules.module import no_p2p_copy
+from ..util.tensor import get_for_device
+
+
+# Synthetic architecture for the MTP head shipped alongside Qwen3.5/3.6 checkpoints.
+#
+# The MTP weights live inside the original (BF16) parent checkpoint under the "mtp." prefix
+# and are *not* part of the main exl3 quant produced by convert.py today. To use this draft:
+#
+#   - Load the target model normally (e.g. an exl3 quant of Qwen3.6-27B with MTP tensors stripped)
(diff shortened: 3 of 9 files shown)


turboderp:
I'm going to look this over tomorrow. But the `rows <= 32` check isn't arbitrary. Without the reconstruct path, prefill would slow down a lot, which might be what you're seeing (though I would expect much more than 33%).

I'm not sure exactly how MTP is implemented for Qwen3, but I'm also currently working on the same for Gemma4, and at least for Gemma4 there should be no impact on prefill. I'll know more tomorrow I guess.

alexhunt7:
I've rebased and reverted the commit for the row count threshold.

Thanks for taking a look.

alexhunt7:
@turboderp Rebased again. I'd love a review if you get the time!

turboderp:
Sorry it took a while to get to this. I had a bunch of other things take priority for a while, but it's merged in the `dev` branch now.

There were a few things that weren't quite right, I think. The MTP model is supposed to use the post-norm hidden state, not the pre-norm state right after the last transformer block. Apparently this was an issue in other frameworks as well and only corrected in the last few days (probably unclear instructions from the Qwen team.) And secondly, prefill wasn't working at all, since the last Transformer block exits early during prefill (no reason to run the last MLP module), it would end up with an empty `export_states` list, skip prefill on the MTP layer and just leave the prompt keys/values uninitialized.

Neither of these would have broken inference or made the draft useless, but they would have had a negative impact on performance.

I did some refactoring as well, moving the initial norms, concatenation and projection into a separate module. This was the easiest way I could find to integrate it all while allowing for quantization. It also allows the MTP model to mostly function like a regular draft model in the generator (with a few other changes to make sure prefix caching doesn't prevent the target model from yielding an output hidden state with which to seed the draft model.) And not overriding `forward()` in the model architecture will make it easier to add autosplit loading, TP inference and more in the future. 

The MTP model is now loaded as a component model rather than a separate architecture, similar to how vision towers are loaded. I think this achieves the most code reuse, and it'll be a cleaner pattern for extending it with more MTP architectures in the future. Probably a lot of the internals in the generator will have to change to accommodate that, but I'm hoping to converge on a neat, unified interface for all of them over time.

Here is a trace-driven benchmark on an RTX 6000 Pro:

| Category                |   Baseline |      MTP (greedy) |   DFlash (greedy) | DFlash : MTP |
|:------------------------|-----------:|------------------:|------------------:|-------------:|
| Agentic, code           |  63.88 t/s | 180.51 t/s, 2.83x | 174.27 t/s, 2.72x |         0.96 |
| Agentic, curl           |  61.42 t/s | 154.40 t/s, 2.51x | 137.04 t/s, 2.23x |         0.88 |
| Coding                  |  68.17 t/s | 187.97 t/s, 2.76x | 224.75 t/s, 3.29x |         1.20 |
| Creative                |  68.29 t/s | 138.22 t/s, 2.02x | 118.24 t/s, 1.73x |         0.86 |
| Creative (reasoning)    |  68.01 t/s | 135.00 t/s, 1.98x | 122.64 t/s, 1.80x |         0.90 |
| Translation             |  66.88 t/s | 139.41 t/s, 2.08x | 100.72 t/s, 1.51x |         0.72 |
| Translation (reasoning) |  66.79 t/s | 162.90 t/s, 2.44x | 147.80 t/s, 2.21x |         0.90 |
| Trivial repetition      |  68.87 t/s | 237.59 t/s, 3.45x | 625.82 t/s, 9.08x |         2.63 |

Overall, it's working great. :cat2: DFlash is a lot faster on a certain type of short one-shot prompt, and a bit better overall in agentic coding tasks. But all things considered I'd say it loses to MTP. Which surprises me a little bit. In a good way, though, since that's a pretty high bar.

The MTP model is now quantized as well, though older EXL3 models won't include it at all. I uploaded a .safetensors file with quantized MTP weights [here](https://huggingface.co/turboderp/Qwen3.6-27B-MTP-exl3) which the loader will automatically scan if it's found in the same directory as the other model weights.

Oh, and I also changed the default draft size to 4, since that just performed better in all my tests.

alexhunt7:
Thank you for merging and fixing those issues. I have encountered an issue when running with multiple GPUs, though. I've opened a PR with a potential fix.
https://github.com/turboderp-org/exllamav3/pull/223

PR #224, opened 2026-06-12
Fix mtp tp

randoentity:
Fixes MTP with TP and includes fix from @alexhunt7 (non-TP multi GPU).

AI use disclosure: the changes were made by LLM. I tested both TP and auto split paths successfully.

See https://github.com/theroyallab/tabbyAPI/pull/424#issuecomment-4668900367 for more details.


randoentity:
All fixed and working at least from 783bfa7.

PR #231, opened 2026-06-20
Fixed key_fused_qkvz and key_fused_ba being overwritten with None

Kira-PH:
Merged the top conditionals for key_fused_qkvz and key_fused_ba into the set of conditionals that follow them, so their values don't get clobbered.

What we had was:
```
if thing:
  set a = thing
  set b = None

if other_thing:
  set a = None
  set b = other_thing
else:
  set a = None
  set b = None
```

The problem is that the `else` overwrites the true condition of the preceding condition.


turboderp:
Thanks. This appears to have snuck in while adding TP support for Qwen3.5

PR #232, opened 2026-06-20
Update pydantic version constraint in requirements

Kira-PH:
Pydantic 2.12.0 adds `pydantic.typing` which conflicts with formatron's use.

formatron has an open ticket about exactly this issue that's been open since October 2025


turboderp:
:+1: I've been going back and forth on this, but it does look like pinning the version is correct for the time being. Will move away from Formatron anyway at some point since it seems to be abandoned.

PR #262, opened 2026-07-28
Fix MTP speculative decoding on multi-GPU splits (fixes #260)

### LINKED ISSUE #260: GDN batched rewind launches all layers' jobs on one device — illegal memory access with MTP on multi-GPU layer split (v1.2.0 regression, fix included)
**Summary.** On a hybrid GDN model loaded across 2 GPUs (plain layer split / autosplit, not TP), speculative decoding with the MTP component crashes with `CUDA error: an illegal memory access was encountered` on the first request. Root-caused to the batched state-rewind dispatch introduced in v1.2.0; a tested patch is included below. While validating the fix I also found a second, independent problem in the recurrent state slot lifecycle under drafting — details at the end.

**Setup:** Qwen3.6-27B EXL3 5.00bpw (MTP tensors embedded), 2× RTX 5060 Ti 16 GB (sm_120, no P2P), native Linux, reproduced on both the v1.2.1 release wheel (cu128/torch 2.9) and a source build of current master (CUDA 13.3, torch 2.11/cu130). Loaded via TabbyAPI (`draft_mode: mtp`, `gpu_split_auto`) and also via a minimal script using `Model.from_config(cfg, component="mtp")` as `draft_model`.

## Bug 1: illegal memory access in batched rewind on multi-GPU splits (fix included)

With `CUDA_LAUNCH_BLOCKING=1` the failure lands in the batched rewind path:

```
GPU assert: an illegal memory access was encountered exllamav3/exllamav3_ext/gdn.cu 1755
```

Greedy and sampled generation crash alike, Q8 and FP16 draft cache alike. v1.1.0 with the identical config does not hit this (its per-layer rewind ran each op under the correct device), so this is a v1.2.0 regression from 6d1f742 ("GDN/Mamba2: Collect state rewind ops into batched kernel").

**Root cause:** `_collect_rewind_jobs()` in `exllamav3/modules/gated_delta_net.py` returns a single `device_index` taken from the first GDN layer, with the assumption that "all layers of one cache live on the same device". Under layer-split loading, a single cache's GDN layers span every device in the split, so `batched_conv_rewind(jobs, device_index)` launches kernels on device 0 whose job structs contain device-1 pointers. First draft rejection → rewind → illegal access. Single-GPU rigs never hit it.

**Fix (tested):** group jobs by each layer's actual device and dispatch per device. After this patch, MTP generation runs correctly on the 2-GPU split (~2.9–3.1× accept-speedup over base decode on agentic/code prompts, consistent with the published MTP numbers). The TP path is unaffected — within a worker all layers share a device, so the grouping degenerates to the previous behavior.

```diff
--- a/exllamav3/modules/gated_delta_net.py
+++ b/exllamav3/modules/gated_delta_net.py
@@ -25,36 +25,38 @@
 def _collect_rewind_jobs(layers, slot: int, last_history: int, num_tokens: int):
-    """Split a batch of recurrent-layer states into (conv_jobs, state_jobs) for the batched
-    rewind kernels, plus a device index (all layers of one cache live on the same device). Only
-    GDNLayerState instances (GDN and Mamba2 alike) are batched; any other recurrent-state type
-    sharing the same cache (e.g. SWA, short-conv) falls back to its own .rewind() call, unchanged."""
-    conv_jobs = []
-    state_jobs = []
-    device_index = None
+    """Split a batch of recurrent-layer states into per-device (conv_jobs, state_jobs) for the
+    batched rewind kernels. With layer-split loading a single cache's GDN layers span multiple
+    devices, so jobs must be grouped by the device each layer's state actually lives on --
+    launching them all under one device index dereferences foreign pointers (illegal memory
+    access on any multi-GPU split). Only GDNLayerState instances (GDN and Mamba2 alike) are
+    batched; any other recurrent-state type sharing the same cache (e.g. SWA, short-conv)
+    falls back to its own .rewind() call, unchanged."""
+    jobs_by_device = {}
     for l in layers:
         if isinstance(l, GDNLayerState):
-            if device_index is None:
-                # l.device may be a plain string ("cuda:0") in some TP contexts rather than a
-                # torch.device, so normalize rather than assume a .index attribute
-                device_index = torch.device(l.device).index
+            # l.device may be a plain string ("cuda:0") in some TP contexts rather than a
+            # torch.device, so normalize rather than assume a .index attribute
+            device_index = torch.device(l.device).index
+            conv_jobs, state_jobs = jobs_by_device.setdefault(device_index, ([], []))
             cj = l.rewind_conv_job(slot, last_history, num_tokens)
             if cj is not None:
                 conv_jobs.append(cj)
             sj = l.rewind_state_job(slot, last_history, num_tokens)
             if sj is not None:
                 state_jobs.append(sj)
         else:
             l.rewind(slot, last_history, num_tokens)
-    return conv_jobs, state_jobs, device_index
+    return jobs_by_device


-def _dispatch_rewind_jobs(conv_jobs, state_jobs, device_index):
-    if conv_jobs:
-        ext.batched_conv_rewind(conv_jobs, device_index)
-    if state_jobs:
-        ext.batched_state_rewind(state_jobs, device_index)
+def _dispatch_rewind_jobs(jobs_by_device):
+    for device_index, (conv_jobs, state_jobs) in jobs_by_device.items():
+        if conv_jobs:
+            ext.batched_conv_rewind(conv_jobs, device_index)
+        if state_jobs:
+            ext.batched_state_rewind(state_jobs, device_index)


 def mp_cache_recurrent_rewind(local_context: dict, cache_id: int, slot: int, last_history, num_tokens):
     recurrent_modules = local_context["recurrent_modules"]
     layers = [module.tp_recurrent_lookup[cache_id] for module in recurrent_modules]
-    conv_jobs, state_jobs, device_index = _collect_rewind_jobs(layers, slot, last_history, num_tokens)
-    _dispatch_rewind_jobs(conv_jobs, state_jobs, device_index)
+    _dispatch_rewind_jobs(_collect_rewind_jobs(layers, slot, last_history, num_tokens))
```

The two call sites (`mp_cache_recurrent_rewind` and `GDNState.rewind`) are updated to the new signatures accordingly. Happy to open this as a PR if preferred.

Possibly related: #245 reports the same crash class under TP + non-greedy on WSL2 — different path, but worth re-testing after this fix. The v1.1.0 `list index out of range` during drafting on cached prompts may also be the pre-batching form of this same lifecycle and appears superseded by the 6d1f742 rewrite once the grouping is corrected.

## Bug 2 (separate, reachable only after fixing bug 1): recurrent state slots exhaust under drafting

With the fix above applied, MTP is stable for repeated requests on the *same* prompt (alloc/release balanced 1:1 in `cache.free_list`, verified with debug logging in `Cache.get_new_state` / `release_state`). But each request whose prompt is not an exact repeat permanently consumes a slot from the pool (`num_slots = max_batch_size`, default 4 for recurrent models). Trace over sequential single-job requests with distinct prompts: four allocations with no interleaved release, one LRU-style burst reclaim of all 4 slots firing once, then `get_new_state` asserting `Cannot create new state: no available slots` on the next request — after which every subsequent request aborts until the process is restarted. Growing prompts (normal multi-turn conversations) behave like distinct prompts, so in practice the generator wedges after ~4-5 turns of real use. Without a draft model the identical request series stays perfectly balanced, so the leak is specific to the draft path.

Raising `max_batch_size` is not a viable workaround on consumer VRAM: each slot costs roughly 1.5–2 GB for this model (48 GDN layers × rollback history for draft verification), so 6/8/16 slots all fail to load alongside the weights on 2×16 GB ("Insufficient VRAM in split for model and cache").

Suggested directions (from reading, not tested): have `get_new_state()` trigger the same reclaim the observed burst used instead of asserting, and/or release a completed job's live state eagerly once it has been stashed to the sysmem RecurrentCache. I can open this as a separate issue if you prefer it tracked independently — happy to provide the full debug trace either way.

dushyant30suthar:
Both fixes from #260, as offered there.

**1. Group batched rewind jobs by each layer's actual device** (`modules/gated_delta_net.py`)

`_collect_rewind_jobs` assumed all of a cache's GDN layers live on one device and launched every rewind job under the first layer's index. With layer-split loading the layers span every device in the split, so `batched_conv_rewind`/`batched_state_rewind` dereferenced foreign pointers — CUDA illegal memory access on the first draft rejection. Jobs are now grouped per device and dispatched under the right guard/stream. The TP path is unaffected: within a worker all layers share a device, so the grouping degenerates to the previous single-entry behavior.

**2. Completion wins over requeue** (`generator/generator.py`)

In the drafting branch of `iterate_gen` the requeue check ran before the EOS check. A job producing its final token while a requeue condition held was requeued instead of completed; the replacement carried a zero-or-negative remaining token budget, could never reach EOS, requeued itself indefinitely, and each orphan held a recurrent state slot — with the default pool of 4, the fifth distinct request failed on `Cannot create new state: no available slots` and the generator wedged. `rq and not eos` restores the intended precedence; non-EOS requeues (output chunking, rollback recovery on long generations) are untouched.

**Validation:** 2× RTX 5060 Ti (sm_120), native Linux, CUDA 12.8 and 13.3, Qwen3.6-27B EXL3 5.0bpw with embedded MTP tensors, via TabbyAPI and via the bare generator API. 30/30 requests across the failure patterns from #260 (max_tokens=1 series, distinct prompts, growing multi-turn conversation), plus full 4k–163k depth sweeps with drafting active on both layer split and `tensor_parallel` — no illegal memory access, no slot exhaustion, draft acceptance in the expected range (~2.5–3.1× accept-speedup on code/agentic prompts).

### DIFF
--- a/exllamav3/generator/generator.py
+++ b/exllamav3/generator/generator.py
@@ -952,7 +952,7 @@ class Generator:
                     # uses the full sequence generated so far as its next prompt. Unconsumed draft positions must
                     # be rejected here so the recurrent state and page positions match the accepted sequence
                     # before the requeue stash.
-                    if len(job.sequences) == 1 and rq:
+                    if len(job.sequences) == 1 and rq and not eos:
                         if draft_tokens is not None:
                             rejected = reject_remainder(job, j, i, batch_states)
                         requeuing_jobs.append(job)
--- a/exllamav3/modules/gated_delta_net.py
+++ b/exllamav3/modules/gated_delta_net.py
@@ -23,19 +23,20 @@ from .attention_fn.bc_attn import MAX_BSZ as _BC_MAX_BSZ, MAX_QLEN as _BC_MAX_QL
 
 
 def _collect_rewind_jobs(layers, slot: int, last_history: int, num_tokens: int):
-    """Split a batch of recurrent-layer states into (conv_jobs, state_jobs) for the batched
-    rewind kernels, plus a device index (all layers of one cache live on the same device). Only
-    GDNLayerState instances (GDN and Mamba2 alike) are batched; any other recurrent-state type
-    sharing the same cache (e.g. SWA, short-conv) falls back to its own .rewind() call, unchanged."""
-    conv_jobs = []
-    state_jobs = []
-    device_index = None
+    """Split a batch of recurrent-layer states into per-device (conv_jobs, state_jobs) for the
+    batched rewind kernels. With layer-split loading a single cache's GDN layers span multiple
+    devices, so jobs must be grouped by the device each layer's state actually lives on --
+    launching them all under one device index dereferences foreign pointers (illegal memory
+    access on any multi-GPU split). Only GDNLayerState instances (GDN and Mamba2 alike) are
+    batched; any other recurrent-state type sharing the same cache (e.g. SWA, short-conv)
+    falls back to its own .rewind() call, unchanged."""
+    jobs_by_device = {}
     for l in layers:
         if isinstance(l, GDNLayerState):
-            if device_index is None:
-                # l.device may be a plain string ("cuda:0") in some TP contexts rather than a
-                # torch.device, so normalize rather than assume a .index attribute
-                device_index = torch.device(l.device).index
+            # l.device may be a plain string ("cuda:0") in some TP contexts rather than a
+            # torch.device, so normalize rather than assume a .index attribute
+            device_index = torch.device(l.device).index
+            conv_jobs, state_jobs = jobs_by_device.setdefault(device_index, ([], []))
             cj = l.rewind_conv_job(slot, last_history, num_tokens)
             if cj is not None:
                 conv_jobs.append(cj)
@@ -44,21 +45,21 @@ def _collect_rewind_jobs(layers, slot: int, last_history: int, num_tokens: int):
                 state_jobs.append(sj)
         else:
             l.rewind(slot, last_history, num_tokens)
-    return conv_jobs, state_jobs, device_index
+    return jobs_by_device
 
 
-def _dispatch_rewind_jobs(conv_jobs, state_jobs, device_index):
-    if conv_jobs:
-        ext.batched_conv_rewind(conv_jobs, device_index)
-    if state_jobs:
-        ext.batched_state_rewind(state_jobs, device_index)
+def _dispatch_rewind_jobs(jobs_by_device):
+    for device_index, (conv_jobs, state_jobs) in jobs_by_device.items():
+        if conv_jobs:
+            ext.batched_conv_rewind(conv_jobs, device_index)
+        if state_jobs:
+            ext.batched_state_rewind(state_jobs, device_index)
 
 
 def mp_cache_recurrent_rewind(local_context: dict, cache_id: int, slot: int, last_history, num_tokens):
     recurrent_modules = local_context["recurrent_modules"]
     layers = [module.tp_recurrent_lookup[cache_id] for module in recurrent_modules]
-    conv_jobs, state_jobs, device_index = _collect_rewind_jobs(layers, slot, last_history, num_tokens)
-    _dispatch_rewind_jobs(conv_jobs, state_jobs, device_index)
+    _dispatch_rewind_jobs(_collect_rewind_jobs(layers, slot, last_history, num_tokens))
 
 
 class GDNState:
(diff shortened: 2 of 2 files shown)


dushyant30suthar:
Both fixes have landed in `dev` (ce77f0a and 1048003) — thanks for taking them, and the `receive_sample()` placement for the requeue/EOS one is the better call since it covers the non-draft path too.

Closing this in favour of those commits. I've been running `dev` equivalents on the 2× RTX 5060 Ti setup with tensor_parallel + MTP for a full day of real agentic traffic since: 787 requests, ~181k tokens generated, 97.6% prefix cache hit rate, no slot exhaustion and no rewind crashes.

PR #263, opened 2026-07-29
Fix int32 offset overflow in the Triton prefill kernel at large geometries

pdcruze:
## Problem

The Triton prefill kernel computes element offsets in int32, which wraps negative once a tensor crosses 2^31 elements. Two paths are affected:

- **Q/out addressing** (`(row * n_q_heads + q_head) * head_dim + d`): overflows past 2^31 elements — e.g. 262,144 query rows at 32 heads × 256 head_dim. The q load then reads far out of bounds and the out store never lands.
- **Split-partial addressing** (`pid_lin * BLOCK_M * head_dim`): overflows when `programs * num_splits * BLOCK_M * head_dim` crosses 2^31, in both the prefill store and the combine-pass loads.

Separately, the auto-split heuristic can misbehave at large program counts: ceil-noise in the cost model can pick a high split count whose fp32 partial buffer runs to tens of GB (e.g. ~17 GB at 64k tokens × 32 heads) for negligible scheduling gain, causing OOM. We hit this evaluating a Gemma3-class model at long context.

## Fix

- Widen the row term (q/out) and `pid_lin` (partials) to int64 in `_paged_attn_prefill_kernel` and `_paged_attn_prefill_combine_kernel`.
- Clamp the auto-picked `num_splits` so the partial buffer stays under 128 MB, and guard the heuristic against `programs == 0`.

## Tests

`tests/test_triton_paged_overflow.py` crosses the 2^31 element-offset boundary in both paths and checks exact expected values, using a NaN sentinel in `out` so a wrapped store is detected. Both tests were verified to fail against the previous kernel and pass with the fix. Note for CI: crossing 2^31 elements inherently requires large allocations, so the tests need ~10–11 GiB of free VRAM and skip below that. A third small test covers the `programs == 0` guard.

### DIFF
--- a/exllamav3/modules/attention_fn/triton_paged.py
+++ b/exllamav3/modules/attention_fn/triton_paged.py
@@ -1384,7 +1384,10 @@ def _paged_attn_prefill_kernel(
     offs_d = tl.arange(0, head_dim)
     valid_row = offs_m < q_len
 
-    q_ptrs = q + (((batch * q_len + offs_m[:, None]) * n_q_heads + q_head) * head_dim + offs_d[None, :])
+    # q/out element offsets reach q_len * n_q_heads * head_dim, past int32 at ~64k rows for
+    # 32-head/256-dim geometry: widen the row term
+    row64 = (batch * q_len + offs_m[:, None]).to(tl.int64)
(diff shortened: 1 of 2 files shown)


turboderp:
Thanks. :+1: Merged into `dev` but I added a gate to keep the int32 indexing for small enough tensors, since int64 indexing showed a small performance cost.

PR #264, opened 2026-07-30
Exclude multimodal-dependent pages from the CPU cache tier

pdcruze:
The CPU tier currently stores any complete content-hashed page, including pages whose K/V depends on multimodal embeddings. Those entries can't do useful work and can do harm:

- Page hashes key on MM token IDs, which are embedding *indices*, not image content. Fresh embeddings always draw new indices from `MMTokenAllocator`, so a re-uploaded identical image never produces a hit — the entries just occupy slots.
- Embeddings deserialized via `MMEmbedding(imp=...)` carry externally assigned indices. If a serving layer reuses an index range for different content, a hit restores K/V from the wrong image — and the tier's retention window is much longer than the GPU pool's, so the aliasing exposure is much larger.
- The dependency propagates: through attention, every page *downstream* of an MM page carries image-dependent K/V under a hash that only commits to token IDs, and the dedup path refreshes recency without replacing bytes. So the whole descendant chain has to be excluded, not just the page containing the MM tokens.

This PR adds `CPUPageCache.mm_tainted()`: a page is rejected if its own tokens contain an MM index or if any ancestor's do, walking ancestry through live VRAM pages and terminating clean at the chain root or at an existing tier entry (entries are only admitted untainted, so their ancestry is pre-verified). Unverifiable chains are rejected conservatively. Verdicts are memoized by page hash — taint is a pure function of the chained hash, so cached verdicts never go stale — keeping an N-page chain teardown O(N) under leaf-first eviction (measured 25 ms for a 1000-page chain, stores included). Rejections are counted in a new `mm_rejects` metric.

Tests included (`tests/test_cpu_cache.py`): taint propagation through text-only descendants, clean-ancestry admission via a tier entry, conservative rejection of a broken chain, a structural assertion that the memo eliminates repeat chain walks, and a store/fetch roundtrip byte-comparison test for the tier itself.

### DIFF
--- a/exllamav3/generator/cpu_cache.py
+++ b/exllamav3/generator/cpu_cache.py
@@ -4,6 +4,7 @@ import threading
 import torch
 from collections import deque
 from ..constants import PAGE_SIZE
+from ..tokenizer.mm_embedding import FIRST_MM_EMBEDDING_INDEX
 
 
 def _align(n: int, a: int) -> int:
@@ -82,12 +83,18 @@ class CPUPageCache:
         self._order_pops = 0
         self._order_rebuild = max(64, self.max_slots // 8)
 
+        # Multimodal taint verdicts by page hash (see mm_tainted). Verdicts are content-addressed and never
+        # stale; the cap only bounds memory, and clearing costs one O(chain) rebuild on the next store
+        self._taint_memo = {}
+        self._taint_memo_cap = 65536
+
         self.metrics = {
             "pushes": 0,        # pages copied to the tier on GPU eviction
             "dedup_hits": 0,    # pushes skipped because the page was already stored
             "restores": 0,      # pages copied back into the GPU cache at allocation
             "evictions": 0,     # tier entries dropped to make room
             "cold_allocs": 0,   # pushes that had to pin a slab synchronously (spare pool was empty)
+            "mm_rejects": 0,    # pushes declined because the page's K/V depends on multimodal embeddings
         }
 
         # Transfers run at PCIe speed, but pinning host memory only manages ~2.5 GB/s and serializes with copy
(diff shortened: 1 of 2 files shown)


turboderp:
I'm not sure what the purpose of this is. Cache pages with multimodal tokens are valid as long as long as the corresponding MMEmbedding is around, and its lifetime is managed by the caller.

There are genuine cases where two different contexts are competing for limited GPU cache and will temporarily have pages swapped to the CPU cache that may or may not contain MM tokens. Once swapped back to GPU memory, the tokens retain their meaning (specific to the corresponding job) as long as the MMEmbedding object is supplied to the Job constructor.

This also allows tokens to persist across multiple jobs, e.g. a context may start with a range of MM tokens and then a back-and-forth conversion can happen where each round is a new job that reuses the *same* MM tokens, not the same image re-encoded with new tokens.

Prematurely trimming out contexts with MM tokens would break that and force jobs to repeat the same deterministic prefill passes multiple times. Whereas not doing so would simply leave the pages in the cache where they will eventually be evicted like any other abandoned context.

Collisions won't happen because the page table and the CPU cache are both tied to the lifetime of the Generator, and all MMEmbedding token ranges are unique and non-overlapping within the process. So there shouldn't be a possible condition where you can encode two different images to the same token IDs without recreating the cache in-between.

pdcruze:
Thanks. That clarifies it, and you're right. I had treated MM token indices as identifiers that could end up referring to different content at different times, but that can't happen. Token ranges are unique within the process and never reused. imp only carries the same identity over to the TP workers, and a follow-up job is meant to be given the same MMEmbedding object rather than a re-encoded image with new tokens. So a page with MM tokens that gets swapped out to the CPU cache and later swapped back still means exactly what it did before, as long as the caller keeps the embedding around (which is the caller's job).

That also means the exclusion would hurt the cases where this matters most: a conversation that starts with an image and continues over several jobs, or two contexts competing for limited GPU cache with MM pages moving between tiers. Those would end up repeating the same deterministic prefill for no benefit. My bad. Withdrawing and closing.

One small piece is unrelated to the MM question, if it's of any use: the store/fetch byte-roundtrip unit test for CPUPageCache. Happy to resubmit just that separately. Otherwise I will withdraw all of this.

Issue #270, opened 2026-07-31
[Feature Request] Add Support for Dry_multiplier and Logit_bias Sampler Parameters


Tester1482:
Hi,

I thought DRY multiplier was supported by exl 3 as it was in exl 2 (as per https://github.com/turboderp-org/exllamav2/compare/affdc0d16c45ac785ccf97278731219260dbc281...c1fed2ed19a720ca0ae6aaa50c378ae350a97df6) but it isn't? After a fresh install of tabby (with exl 1.1) I got the message `2026-07-30 23:57:03.551 WARNING:  Ignoring sampler params not supported by the 
exllamav3 backend: dry_multiplier, logit_bias` Any chance this could be re-implemented?

Thanks again for all your continued hard work and support, it's absolutely phenomenal what you have accomplished.

Ph0rk0z:
I remember it being supported inside actual EXL3.. sounds like problem is on tabby end for EX3 specifically?

sashko-zakharchuk:
Both of these were missing on the exllamav3 side, not only in tabby.

`logit_bias` is in as of v1.4.1 (#274), and the tabby half is theroyallab/tabbyAPI#445, merged
on 8 August, so the warning clears once both are current.

DRY is still missing (no `dry_multiplier` anywhere in the tree). #278 adds it, still open.

PR #274, opened 2026-08-02
Add logit_bias sampler parameter

sashko-zakharchuk:
Adds the OAI-style `logit_bias` sampler parameter as `SS_LogitBias`. This is the exllamav3-side building block for #270 — the warning the issue quotes (`Ignoring sampler params not supported by the exllamav3 backend: dry_multiplier, logit_bias`) is emitted by TabbyAPI, which also needs a small change to pass the parameter through; I'll open that companion PR against TabbyAPI. (`dry_multiplier` is not included here.)

`SS_LogitBias` adds a fixed per-token bias to the logits, modeled on `SS_RepP`: it runs among the first steps, before the logits are transformed. A bias of `-inf` bans a token; zero, `nan`, and values that overflow float32 to `+inf` are ignored. The bias vector is cached per (vocab size, device) like `TokenMask`, a leading bias step participates in the fused-sampler fast path like `SS_BanTokens`, and it no-ops when empty. Wired into `ComboSampler` as `logit_bias=...`.

Note: exllamav2 applies its token bias *after* the (multiplicative) repetition penalty; this applies `logit_bias` to the raw logits, matching the OAI meaning of the parameter.

Three cases added to `tests/test_sampler.py` (additive with OOB/zero/nan/+inf handling across a batch; `-inf` ban; a chain that exercises the `SS.LOGITS` branch) plus a fused-collapse assertion. The sampler test suite passes (140/140).

### DIFF
--- a/exllamav3/generator/sampler/__init__.py
+++ b/exllamav3/generator/sampler/__init__.py
@@ -17,6 +17,7 @@ from .custom import (
     SS_PresFreqP,
     SS_AdaptiveP,
     SS_BanTokens,
+    SS_LogitBias,
     SS_XTC,
     xtc_default_protected_token_ids,
 )
--- a/exllamav3/generator/sampler/custom.py
+++ b/exllamav3/generator/sampler/custom.py
@@ -577,6 +577,63 @@ class SS_BanTokens(SS_Base):
         return None
 
 
+class LogitBias:
+    """
+    Additive per-token logit bias vector with per (vocab size, device) caching. Token IDs outside
+    the vocabulary are ignored.
+    """
+    def __init__(self, logit_bias: dict[int, float]):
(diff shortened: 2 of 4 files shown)


Ph0rk0z:
So does that mean banned strings also do not work? Plain logit bias is ehhh

sashko-zakharchuk:
Banned strings work on the exllamav3 backend already: TabbyAPI passes `banned_strings` through to the generator, and exllamav3 handles it natively in the Job (string matching with hold-and-rewind), so it never needed a sampler stage. `logit_bias` is different: it needs a per-token bias on the logits, which is what this PR adds as a sampler step. The TabbyAPI side that stops ignoring the parameter is theroyallab/tabbyAPI#445.

PR #276, opened 2026-08-04
DFlash: read block_size from dflash_config when missing at top level

CCobaltDev:
Some DFlash models include `block_size` in `dflash_config` instead of at the top-level, as seen in [z-lab/Qwen3.6-35B-A3B-DFlash](https://huggingface.co/z-lab/Qwen3.6-35B-A3B-DFlash)


turboderp:
👍

PR #246, opened 2026-07-15
[kernel] Persistent route-packed (route_block x n_tile) schedule for exl3 fused MoE (sm120)

brandonmmusic-max:
# PR (draft) — A1 route-packed re-tile for the EXL3 fused MoE kernel (sm120)

**Target:** `turboderp-org/exllamav3` (based on v0.0.43).
**Independent** of the nvfp4 KV work — a standalone sm120 MoE-kernel speedup.

## Title
`[kernel] Persistent route-packed (route_block × n_tile) schedule for exl3 fused MoE on Blackwell (sm120)`

## Motivation
The EXL3 fused grouped-MoE kernel underuses consumer Blackwell (sm120):
- `quant/exl3_moe_common.cuh`: `#define MOE_SMS_PER_EXPERT 8` → concurrency 188/8 = 23, **4 SMs
  idle**; and smem is hardcoded `90*1024` ("CC 8.6") while **sm120 allows ~99 KiB opt-in**.
- The fixed 8-SM-per-expert grouping leaves the grouped GEMM tail-heavy and under-tiled.

## The change
Port the b12x W4A16 persistent schedule idea onto the exl3 grouped shapes: a **persistent
`(route_block × n_tile)` grid** that keeps all 188 SMs busy and split-Ks the grouped GEMM,
replacing the 8-SM-per-expert fixed grouping; use the full sm120 smem opt-in; keep the trellis
codebook dequant (`codebook.cuh`, `exl3_dq.cuh`) and the fp16/bf16 `mma.m16n8k16` accumulate path
**byte-identical** (same numerics). Added as a new entry point `exl3_moe_fused_retile` **alongside**
the existing `exl3_moe`/`exl3_moe_fused` for A/B; opt-in.

Files (diff vs v0.0.43): `quant/exl3_moe.cu` (+447), `quant/exl3_gemm_inner.cuh` (+373),
`quant/exl3_moe_kernel.cuh` (+158), `quant/exl3_moe_common.cuh` (+46).

## Result
**Prefill +7–8%** on GLM-5.2's 192-expert trellis-3.0 tail (4× RTX PRO 6000 Blackwell, TP4).
Numerics unchanged (trellis dequant + MMA path untouched); shipped and serving coherently since
`verdictai/vllm-glm52-tr3-hybrid:v3`. The re-tiled path is additive/opt-in, so the existing
`exl3_moe`/`exl3_moe_fused` remain byte-identical for fallback.  

---
[...]

PR #284, opened 2026-08-13
MoE: Add fused additive EXL3 kernels

malaiwah:
## Summary

Add a graph-capturable additive EXL3 MoE path for compressed residual ("cartridge") stages.

- extend `exl3_gemm_inner` with FP32 accumulation and output scaling
- add `exl3_moe_additive` for callers that already have route metadata
- add `exl3_moe_additive_fused` for stable route packing plus base/residual execution in one extension call
- export `EXL3_MOE_ADDITIVE_ABI_VERSION = 1` for exact consumer compatibility checks
- tile oversized expert route spans within the fixed scratch workspace
- support stage-major trellis pointer, scale, and bitrate tables with zero-scale sparse entries
- validate CUDA device, contiguity, dtype, shape, workspace, and routing contracts before device-side mutation
- preserve the existing `exl3_moe` ABI and legacy overflow fallback behavior

## API contract

Consumers must require an exact supported `EXL3_MOE_ADDITIVE_ABI_VERSION`; symbol presence or callable arity is not a compatibility contract. The version is bumped for incompatible signature, pointer-layout, routing-workspace, residual-encoding, or overflow-semantics changes.

Additive residual trellises use the MCG codebook and reuse each base projection's `suh`/`svh` rotations. Their K metadata must be validated by the caller before graph capture. A zero scale skips that projection's GEMM.

The `int64` pointer tables do not retain their pointees. Callers must keep all referenced trellis and rotation allocations alive, unmoved, and on the same device through asynchronous completion and for the lifetime of a captured graph.

For additive calls, `num_active` is only a launch-size hint and counts every nonempty expert. Oversized experts are processed in workspace-sized tiles, so `rows * topk` scratch is not required.

## Relationship to #246

The fused route histogram, scan, and stable-packing pipeline is adapted from @brandonmmusic-max's draft #246 (commit `704aefd`) and rebased onto the current ticket-scheduled MoE kernel. This PR does not include #246's SM120 retile; it adds additive residual stages. Brandon is credited in the squashed commit.

## Companion work

- runtime consumer: [local-inference-lab/vllm#299](https://github.com/local-inference-lab/vllm/pull/299)
- cartridge producer: [malaiwah/progressive-tensors#41](https://github.com/malaiwah/progressive-tensors/pull/41)

This PR is intentionally limited to the extension compute/API layer; it does not define the on-disk cartridge manifest. Consumers remain responsible for validating artifact version, extension ABI, codebook, K values, coverage, and pointer lifetime before invoking the extension.

## Validation

Direct extension validation on NVIDIA L4 (SM89), CUDA 12.6, and PyTorch 2.6:

- full `setup.py build_ext --inplace` completed successfully
- `python -m pytest -q tests/test_exl3_moe_additive.py`: **14 passed**
- covers int32/int64 route IDs; FP16/BF16/FP32 route weights; stable packing; exactly `capacity + 1` and multi-tile overflow; the downstream `topk=2`, `capacity=rows` case; zero-scale/null sparse stages; and preserved legacy overflow behavior

The ABI-version follow-up is a header/binding/test-only change. `git diff --check` and Python syntax validation pass; the CUDA behavior is unchanged from the L4-tested commit.

Earlier downstream prototype validation on RTX 5090 (SM120), CUDA 12.8, and PyTorch 2.8 reported successful CUDA-graph capture and base -> cartridge -> base transitions. Restored base logprobs matched exactly (`max_abs=0.0`); packed-vs-dense cartridge output had mean KL `1.30506e-4`, max absolute logit delta `0.11122`, mean absolute logit delta `0.00910`, and identical argmax on sampled prompts.

AI assistance was used. I reviewed the source changes and validation results.

### DIFF
--- a/exllamav3/exllamav3_ext/bindings.cpp
+++ b/exllamav3/exllamav3_ext/bindings.cpp
@@ -208,6 +208,13 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
     m.def("blocksparse_mlp_routing", &blocksparse_mlp_routing, "blocksparse_mlp_routing");
     m.def("exl3_moe_max_concurrency", &exl3_moe_max_concurrency, "exl3_moe_max_concurrency");
     m.def("exl3_moe", &exl3_moe, "exl3_moe");
+    m.attr("EXL3_MOE_ADDITIVE_ABI_VERSION") = EXL3_MOE_ADDITIVE_ABI_VERSION;
+    m.def("exl3_moe_additive", &exl3_moe_additive, "exl3_moe_additive");
+    m.def(
+        "exl3_moe_additive_fused",
+        &exl3_moe_additive_fused,
+        "exl3_moe_additive_fused"
+    );
 
     m.def("bighead_attn", &bighead_attn, "bighead_attn");
     m.def("bighead_attn_paged", &bighead_attn_paged, "bighead_attn_paged");
--- a/exllamav3/exllamav3_ext/quant/exl3_gemm_inner.cuh
+++ b/exllamav3/exllamav3_ext/quant/exl3_gemm_inner.cuh
@@ -30,7 +30,9 @@ void exl3_gemm_kernel_inner
     const int size_k,
     const int size_n,
     int* __restrict__ locks,
-    const half* post_scale
+    const half* post_scale,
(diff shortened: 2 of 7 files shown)


brandonmmusic-max:
Thank you for the attribution

Issue #50, opened 2025-06-09
RuntimeError: Insufficient VRAM in split for model and cache despite remaining 1GB VRAM (exllamav3 + tabbyAPI)


meigami0:
I encountered a VRAM allocation issue while testing exllamav3 quantized models with TabbyAPI. Even with ~1GB VRAM remaining, the system throws an error: `RuntimeError: Insufficient VRAM in split for model and cache`. Below are test results for two models on 24564MiB VRAM (RTX4090):  

#### Model 1: **Nemotron-Super-49B** (3.0bpw, `cache-mode FP16`)  
- 2k context: 19162MiB used  
- 12k context: 21342MiB used  
- 20k context: 23094MiB used  
- 21k context: 23194MiB used  
- 22k context: OOM  

#### Model 2: **gemma3-27b** (5.0bpw, `cache-mode Q6`)  
- 2k context: 17404MiB used  
- 12k context: 19404MiB used  
- 22k context: 21660MiB used  
- 30k context: 23238MiB used  
- 31k context: OOM  

Environment: WSL Ubuntu+Python12+torch2.7.0+cu128+exllamav3-0.0.3+tabbyAPI default config

**Questions:**  
- Is this behavior expected under exllamav3’s current memory management design?  
- Are there recommended configurations or workarounds to better utilize the remaining VRAM?

turboderp:
A couple of likely contributing reasons:

Momentary VRAM allocations can be so quick that they don't show up in a monitoring utility which might be polling, say, once per second. So if you see:
     
- 20k context uses 23.1 GB, and
- 21k context uses 23.2 GB, so you might expect
- 22k context uses 23.3 GB, but..

... this might not be accounting for 0.75 GB of temporary storage for some large operation that's too fast to notice.

The reason you might get this OoM error while loading is that the model performs a dummy forward pass as it shovels tensors onto the GPU, to determine precisely where the overflow occurs and thus where it should split the model in a multi-GPU setup.

As for that, by default a small buffer amount is reserved on each GPU to keep things stable, especially for the first GPU which might also be running a desktop environment. You can change the size of this reserve in Tabby's config, or if you're calling ExLlama directly, there are arguments to the `load()` and `load_gen()` functions.

[...]

PR #293, opened 2026-08-18
Loader: Account for already-reserved memory in autosplit reserve

DenysAshikhin:
TL:DR
Exl3 is reserving (double-counting) VRAM reserve on model load when multiple models are loaded (i.e. main model + vision + mtp). Currently, I am doing a 4.6 bpw 3.8 27B and can get about 125k context (at q8/q8). Going to 130, it complains OOM. Making this fix lets me go higher (~160k) and fill full context without crashes.
More nuanced (explanation below):
----------
`set_memory_fraction_reserve()` derives the per-process cap from `torch.cuda.mem_get_info()`, which reports memory free after whatever this process has already reserved. `torch.cuda.set_per_process_memory_fraction()` then applies that as a limit on the process's cumulative reserved bytes, so memory allocated by an earlier load in the same process is subtracted from the budget twice.

This affects any caller loading more than one `Model` per process — the pattern in `examples/generator.py` (draft + main) and `examples/multimodal.py` (vision + main), and how TabbyAPI loads a vision tower, an MTP draft and the main model. `_load_autosplit()` installs the cap once per load, so by the third component the ceiling has collapsed by the size of the first two.

Measured on a 24 GB RTX 4090 loading a 27B EXL3 model with a vision tower and MTP drafting at 130k context:

| stage  | already_reserved | cap installed | cap with this change |
| ------ | ---------------: | ------------: | -------------------: |
| vision |            0 MiB |     22940 MiB |            22940 MiB |
| draft  |         1184 MiB |     21646 MiB |            22830 MiB |
| main   |         1568 MiB |     21262 MiB |            22830 MiB |

The main model then fails with `Insufficient VRAM in split for model and cache` while 1678 MiB is still free on the device. Adding the already-reserved bytes back holds the cap steady at 22830 MiB across all three stages and the load succeeds.

Single-component loads are unaffected beyond the few MiB `touch_device()` reserves.

Relates to #50 — this is one cause of that report (the multi-component case); the single-component numbers there are not explained by this change.

### DIFF
--- a/exllamav3/util/memory.py
+++ b/exllamav3/util/memory.py
@@ -30,8 +30,13 @@ def set_memory_fraction_reserve(
 ):
     touch_device(device)
     free, total = torch.cuda.mem_get_info(device)
-    fraction = (free - reserve) / total
-    fraction = max(0.01, fraction)
+    # mem_get_info reports memory free *after* whatever this process has already reserved, but
+    # set_per_process_memory_fraction limits the process's *cumulative* reserved bytes. Add the
+    # current reservation back, or memory held by an earlier load in the same process (a draft
+    # model, a vision tower) is subtracted from the budget twice.
+    current = torch.cuda.memory_reserved(device)
+    fraction = (current + free - reserve) / total
+    fraction = min(1.0, max(0.01, fraction))
     torch.cuda.set_per_process_memory_fraction(fraction, device = device)
 
 


DenysAshikhin:
https://github.com/turboderp-org/exllamav3/issues/50
Also probably same issue as this?

PR #298, opened 2026-08-20
Fix GDN optimizer targets for measurement levels

malaiwah:
## Problem

`conversion/measure_model.py` records a group only when its recursive target tree reaches the requested depth. `GatedDeltaNet.optimizer_targets()` currently returns one nesting level less than `Attention`, and it never includes `o_proj`.

For split Qwen GDN layers this has two concrete effects:

- `measure_model.py -l 3` emits **no GDN groups at all**;
- the default `-l 2` measures the input projections but always omits GDN `o_proj`.

On dense Qwen3.5/3.8 topology, unchanged `-l 3` therefore covers only 256 of 400 quantized body linears and silently omits all 144 GDN linears.

## Why important, what is the impact?

It invalidates `-l 3` as a whole-model allocator:                                                                                                                                                                                                    - 36% of body linears are invisible: 144/400 modules, about 5.54B weights or roughly 23% of quantized body parameters.
- The optimizer gets no propagated KLD sensitivity for GDN `in_proj_qkv`, `in_proj_z`, or `out_proj`. Those modules are either pinned accidentally or dangerously underpriced, biasing bytes toward MLP/full-attention.     
- Any resulting K-map, Pareto estimate, or comparison against ResComp would be incomplete and potentially fidelity-worse despite looking optimal.

## Fix

Match the target-tree depth already used by full attention:

- split layout: QKV, Z and output are three level-3 groups;
- fused layout: QKVZ and output are two level-3 groups;
- B/A remain excluded because they are not quantized targets.

At level 2, the same tree keeps all split GDN projections in one combined group, now including output.

## Regression

`tests/test_optimizer_targets.py` executes the actual `optimizer_targets` method extracted from the source AST, so it needs no model weights, GPU or compiled extension. It asserts split level-3 inputs/output, split level-2 combined coverage, and fused level-3 input/output.

On the Qwen 3.8 model, it verifies that level 3 yields 320 groups covering all 400 body linears exactly once: GDN 144 groups, full attention 48, MLP 128, while retaining k+v and gate+up grouping.

Tested on Python 3.12 by executing all three new regression functions: `3 optimizer-target tests passed`. The companion verifier also rejects the pinned unpatched blob and passes this patched blob.

AI assistance was used to inspect the grouping tree and draft the change; the patch, target counts and regression were reviewed and executed as described.

### DIFF
--- a/exllamav3/modules/gated_delta_net.py
+++ b/exllamav3/modules/gated_delta_net.py
@@ -523,13 +523,17 @@ class GatedDeltaNet(Module):
     @override
     def optimizer_targets(self):
         if self.qkvz_proj is not None:
-            return [[self.qkvz_proj.optimizer_targets()]]
+            return [[
+                self.qkvz_proj.optimizer_targets(),
+                self.o_proj.optimizer_targets(),
+            ]]
 
         targets = []
         if self.qkv_proj is not None:
-            targets += self.qkv_proj.optimizer_targets()
+            targets.append(self.qkv_proj.optimizer_targets())
         if self.z_proj is not None:
-            targets += self.z_proj.optimizer_targets()
+            targets.append(self.z_proj.optimizer_targets())
+        targets.append(self.o_proj.optimizer_targets())
         return [targets]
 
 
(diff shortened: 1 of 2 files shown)


malaiwah:
Added the architecture-level regression requested by review at `a236a3e39fcb3343f369e5bb240587ad24730562`. It executes the checked-in Qwen layer schedule and production optimizer-target methods through a CPU-only fixture, proving 144 GDN + 48 full-attention + 128 MLP = 320 distinct groups and 400 distinct linears with no duplicate target identity. The fixture is also composable with the fused-QKV attention surface. Focused result: **4 passed**.

Issue #297, opened 2026-08-20
[Feature] Emit a versioned uniform fused-QKV EXL3 object for Qwen full-attention blocks


malaiwah:
## Summary

Qwen3.8 full-attention blocks are currently serialized as three separate EXL3 logical modules (`q_proj`, `k_proj`, `v_proj`). That forces a serving consumer to launch three GEMMs and concatenate their outputs. This feature request is for an **optional, versioned, uniform-bit fused-QKV producer format** emitted directly by conversion from the BF16 q/k/v weights.

This is deliberately **not** a mixed-width tuple request. The first implementable topology uses one K/codebook/scale mode for the complete fused matrix.

## Producer contract

For each eligible full-attention block:

1. concatenate the BF16-derived q/k/v matrices in a frozen, declared order;
2. calibrate and quantize that fused matrix as one EXL3 object, rather than concatenating three independently quantized payloads;
3. serialize one versioned logical tensor with explicit q/k/v split metadata, original shapes, K, codebook markers and scale mode;
4. make the safetensors index/config/finalizer able to distinguish split and fused topology;
5. reject partial/mismatched q/k/v inputs and unsupported heterogeneous widths;
6. reconstruct deterministically and expose enough metadata for a serving consumer to issue one launch and split views without load-time re-encoding.

The split topology remains the default/control. A fused option needs new direct marginals because split-module scores do not transfer to a jointly quantized matrix.

## Why this is worth a producer format

On the current Qwen3.8-27B/SM120 stack, measured split decode kernels are approximately 32.1 + 27.8 + 28.1 = **88 microseconds** for q/k/v under B12X. A one-launch fused estimate is about 36 microseconds / roughly 5% decode, but that number is **modeled**, not an integrated result. Evidence: https://github.com/malaiwah/qwen38-27b-exl3/blob/main/receipts/kernel-gap-gemm-bandwidth.json and https://github.com/malaiwah/qwen38-27b-exl3/blob/main/docs/47-kernel-gap-analysis.md.

The current hydrated manifest confirms 64 separate full-attention EXL3 modules (q/k/v/o across 16 layers): https://github.com/malaiwah/qwen38-27b-exl3/blob/main/receipts/hydrated-quantization-manifest.json.

## Acceptance boundary

[...]

PR #299, opened 2026-08-20
PoC: per-layer split or fused-uniform QKV topology

malaiwah:
## Purpose

Proof of concept for #297: allow one checkpoint to choose QKV topology **per full-attention block**.

- `split`: existing q/k/v EXL3 objects, independent K/codebook/scale choices, preserving mixed precision and the serving consumer's three-launch route.
- `fused_uniform`: concatenate the BF16 q/k/v source weights in frozen q,k,v order, quantize that matrix jointly at one common K/codebook/scale, and serialize one qkv payload for a one-launch consumer.

This is the practical hybrid requested by the downstream allocator: retain split widths where q/v sensitivity matters; use uniform fusion only where a new fused direct marginal shows no decision-relevant fidelity loss. A heterogeneous-width one-launch tuple remains future kernel/format work.

## Implementation

Adds versioned `exl3_qkv_topology/1` metadata and a deterministic converter plan:

- complete split-default rows with per-layer fused overrides;
- direct BF16 q/k/v loading and exact q,k,v-order concatenation;
- one joint quantization call retaining the input calibration qmap;
- mixed split/fused Attention construction and runtime routing;
- config, safetensors/index and tensor-storage logical reconstruction metadata;
- fail-closed common-K/codebook/scale, unknown/duplicate declaration, missing/duplicate payload and component checks;
- fused TP explicitly rejected by the PoC rather than silently mis-sharded.

No load-time re-encoding and no duplicate split+fused payload for one block. Split remains the default.

## Verification

The companion source verifier pins the patch/base/patched identities and mixed-schema invariants. Focused tests were executed in the immutable CUDA 13.2/r34 environment after compiling the pinned source extension:

```text
6 passed, 1 warning in 1.89s
```

Coverage includes mixed maps/runtime construction, BF16 concatenation/split bit identity, common-K/codebook/scale enforcement, unknown/duplicate declarations, and missing/duplicate payloads.

A consumer PoC is open at local-inference-lab/vllm#454. The ~88 us split versus ~36 us fused figure remains **modeled**; this PR enables the real block/end-logit/graph/TG experiment and makes no performance claim.

AI assistance was used to implement and review the PoC; the patch identity, extension build and focused tests were independently executed as described.

### DIFF
--- a/exllamav3/conversion/compile.py
+++ b/exllamav3/conversion/compile.py
@@ -7,6 +7,7 @@ from ..loader.safetensors_alt import save_file
 from ..util.memory import free_mem
 from ..modules import Module, Linear
 from .quant_config import update_config, create_quantization_config_json
+from .qkv_topology import canonical_json, validate_payload_index
 
 def tsize(t):
     return t.nelement() * t.element_size()
@@ -26,6 +27,12 @@ def compile_model(args, model, config, tokenizer, mtp_model = None, vision_model
         work_dir = args["work_dir"]
         qtensors_dir = os.path.join(work_dir, "qtensors")
         qtensors_stc = SafetensorsCollection(qtensors_dir, tensor_name_fixes = config.get_tensor_name_fixes())
+    qkv_topology = args.get("exl3_qkv_topology")
+    qkv_safetensors_metadata = (
+        {"exl3_qkv_topology": canonical_json(qkv_topology)}
+        if qkv_topology is not None else
+        None
+    )
 
     # Prepare output directory
     if not os.path.exists(out_dir):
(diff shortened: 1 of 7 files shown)


malaiwah:
Adversarial review follow-up is now in `3777614a5af305d8e2db4bab87f940983e9c71e0`.

- No-plan conversion now short-circuits topology handling and preserves existing behavior, including ordinary 3inst conversions.
- Opted-in topology handles Qwen3.5/3.8 interleaved Q/G with the real doubled q projection width (`[12288,1024,1024]` for the target).
- Schema rows/completeness are scoped to the primary text decoder; vision and MTP stay on their established payloads.
- The producer's opted-in schema domain now matches the consumer: K3..K8 and mcg/mul1, with lexical full-path ordering.
- Added realistic interleaved, side-model exclusion, no-plan, and domain-boundary tests.

Focused result on the pinned GG test image: **21 passed**.

malaiwah:
Actual layer-3 producer pilot found a remaining gated-Qwen execution bug: the fused path serialized the correct 14,336-wide QG/K/V payload but `Attention.project_qkv` still split only 6,144 Q values. `0f92b6f4` now splits the complete 12,288-wide interleaved Q/G component, deinterleaves it through the established fast/fallback paths, and returns the 24×256 query plus 6,144 gate values. Added a forward regression over the real 24/4/256 geometry. Focused producer tests: **22 passed**; combined producer+GDN source tests: **26 passed**.

malaiwah:
The next actual pilot reached the extension fast path and exposed one more concrete contract: `torch.split` returns a non-contiguous Q/G view, while `ext.deinterleave_qg` requires contiguous input. `dce9cfa` materializes only that fused Q/G slice before the existing deinterleave kernel and adds a half-precision fast-path regression that asserts contiguity. Producer tests: **23 passed**; combined producer+GDN tests: **27 passed**.

PR #303, opened 2026-08-22
Speed up Qwen MTP decoding by 22% with a selected vocabulary head

keldenl:
## Result

**A selected 64K draft head increased mean Qwen3.8-27B MTP decode throughput by 21.9%, from 36.58 to 44.58 t/s, on an RTX 4060 Ti 16 GB.** The target still verifies every proposal with its full 248K-token head. The fast path adds about 0.55 GiB of allocation and still fits a configured 131,072-token Q4/Q4 cache.

| Configuration | Code (t/s) | Reasoning (t/s) | Prose (t/s) | Structured (t/s) | Mean (t/s) |
| --- | ---: | ---: | ---: | ---: | ---: |
| Full 248K head, MTP-4 | 30.354 | 41.154 | 31.637 | 43.191 | 36.584 |
| Selected 64K head, FP8 embeddings, MTP-4 | 38.077 | 51.815 | 38.030 | 50.392 | **44.579** |

The selected-head MTP-4 profile also beat the full-head MTP-3 control by 16.8% (44.579 vs. 38.159 t/s).

Benchmark setup: Qwen3.8-27B EXL3 3.00 bpw, Q4 target and MTP KV, batch size 1, argmax, thinking enabled, and four fixed 251-token code/reasoning/prose/structured generations. The branch uses `dev` commit `abf4911` and a CUDA 12.8 extension built for SM 8.9.

## Implementation

Qwen MTP runs the target output projection for each greedy draft step. Qwen3.8 has 248,320 vocabulary rows, so that projection consumes a large share of draft time.

This opt-in EXL3 path:

- builds a smaller draft-only output head from selected vocabulary groups and keeps the matching input embeddings on the GPU;
- chains intermediate MTP token IDs on the GPU, with one CPU readback after each draft block;
- uses the full target head for verification and composes with the confidence-calibrated dynamic-draft pipeline on `dev`.

Unconfigured models keep the existing MTP path.

EXL3 applies its output Hadamard transform in aligned 128-token groups. The implementation accepts complete groups of eight packed blocks; arbitrary block selection or reordering would produce invalid logits. `util/build_mtp_hot_blocks.py` ranks groups from unrestricted MTP proposal traces, then corpus frequency, and retains special-token groups.

## Correctness and tradeoffs

The selected head proposes draft tokens. The target's original full-vocabulary head verifies each proposal. Target weights, target quantization, and speculative-decoding correctness remain unchanged. A poor domain map can lower acceptance and erase the speed gain.

Full-head diagnostics on the rebased `dev` branch found the full-head winner inside the selected vocabulary for 68 of 72 proposals. The selected and full heads matched on all 68 of those proposals. An earlier larger pass matched 582 of 582 proposals under the same condition.

At a configured 131,072-token Q4/Q4 cache:

| Configuration | Allocated | Reserved |
| --- | ---: | ---: |
| Full head, MTP-4 | 12.819 GiB | 13.789 GiB |
| Selected 64K + FP8 embeddings, MTP-4 | 13.369 GiB | 14.102 GiB |

Long contexts shift more time into attention and KV work. MTP-3 beat MTP-4 at high occupancy on this GPU, so operators should benchmark draft depth and context instead of treating the measured 64K/MTP-4 profile as a universal preset.

## Configuration

`MTPHotVocabConfig` configures the MTP component. Standard model initialization exposes:

- `--mtp_hot_blocks`
- `--mtp_hot_embedding_dtype {fp16,fp8}`
- `--mtp_hot_validate`

Environment-variable equivalents support servers that cannot forward component configuration yet. See `doc/mtp_hot_vocab.md`.

Current scope: Qwen3.5/3.6 embedded MTP, an EXL3 target `lm_head`, and single-GPU layer split.

## Validation

- `pytest -q tests/test_mtp_hot_vocab.py`: 9 tests passed
- source `compileall` and `git diff --check`
- real-model selected/full-head numerical checks
- balanced MTP-3/MTP-4 controls and an 85,952-token real-source smoke test

### DIFF
--- a/doc/env_vars.md
+++ b/doc/env_vars.md
@@ -426,3 +426,13 @@ kernel with the kernel name, source line and bad index instead of corrupting mem
 faulting asynchronously downstream. Debug tool for paged-pool issues; significant JIT
 overhead (forces Triton debug mode globally), leave unset in production. AOT/BC graph
 kernels are unaffected (compiled with asserts off).
+
+## MTP hot-vocabulary drafting
+
+See [Qwen MTP hot vocabulary](mtp_hot_vocab.md) for constraints and map generation.
+
+| Variable | Description |
+| --- | --- |
+| `EXL3_MTP_HOT_BLOCKS` | Packed EXL3 block map for a reduced Qwen MTP draft head. |
+| `EXL3_MTP_HOT_EMBED_DTYPE` | Copied draft-embedding dtype: `fp16` (default) or `fp8`. |
+| `EXL3_MTP_VALIDATE_SUBHEAD` | Nonzero compares subset and full draft-head argmax; diagnostics only. |
--- a/doc/mtp_hot_vocab.md
+++ b/doc/mtp_hot_vocab.md
@@ -1,0 +1,77 @@
+# Qwen MTP hot vocabulary
+
+Qwen3.5/3.6 MTP evaluates the target output projection for each draft step. For models with a
+large vocabulary, `MTPHotVocabConfig` can build a smaller EXL3 output head from selected
+128-token groups and copy the matching input embeddings to the GPU.
+
+The reduced head proposes greedy draft tokens. The target runs its full output head and verifies
+each proposal, as in standard speculative decoding. Vocabulary selection can change acceptance
+and speed. Target verification preserves the target distribution.
+
+## Constraints
+
+- Qwen3.5/3.6 with its embedded MTP component
+- EXL3 target `lm_head`
+- layer-split inference on one GPU
+- complete 128-token EXL3 Hadamard groups
+- enough VRAM for the reduced head and copied embeddings
+
+The existing MTP path remains the default.
+
+## Build a block map
+
+Use text from the expected workload and, when available, proposal IDs from unrestricted MTP:
+
+```bash
+python util/build_mtp_hot_blocks.py \
+    -m /models/qwen-exl3 \
+    -c /repos/project \
+    --draft_ids draft-proposals.txt \
+    -b 4096 \
+    -o mtp-hot-blocks.txt
+```
+
+`-b 4096` selects 65,536 tokens. The script ranks groups by proposal frequency, then corpus
+frequency, and retains groups that contain special tokens. Maps depend on the model tokenizer
+and workload.
+
+Proposal files contain one token ID per line. Construct `Generator` with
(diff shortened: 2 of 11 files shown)


turboderp:
Do you have any statistics for how many of the tokens in each 128-wide block end up mattering? I'm wondering if it would be cleaner to just create a dedicated hot_head tensor while quantizing with just the hottest tokens rather than the blocks that happen to contain the hottest tokens. This could also be kept at a lower bitrate to further reduce latency with a likely negligible effect on the draft.

This seems actually really interesting, and something that maybe rather should be a first-class feature.

keldenl:
Yeah, I checked the 2,523 unrestricted MTP proposals used for the map. They contained 650 unique token IDs across 253 128-token groups. The median group had 1 token appear, the mean was 2.57, p90 was 5, and the max was 33. Of the 512 groups in the 64K head, 259 had no hits in that trace. 52 token IDs covered half the proposals, and 398 covered 90%.

So yeah, that seems to support what you’re suggesting! I’m going to try creating the hot head during quantization from the individual token rows, then test a few head sizes and lower bitrates to see what it does to speed and acceptance. If that works well, I agree it would make more sense as a first-class feature.

keldenl:
Ok I just tried this, I rebuilt the head from individual token rows and quantized it separately.

At 4 bpw, it actually matched the full head whenever the winning token was included! The 8K head took 0.032 ms, 16K took 0.165 ms, and 32K took 0.329 ms, which is MUCH faster compared with 0.932 ms for the current 64K grouped head.

Results varied though with acceptance because my proposal trace is probably still too small, so I'm looking to do broader testing before jumping into the converter and model format.

sorasoras:
> Ok I just tried this, I rebuilt the head from individual token rows and quantized it separately.
> 
> At 4 bpw, it actually matched the full head whenever the winning token was included! The 8K head took 0.032 ms, 16K took 0.165 ms, and 32K took 0.329 ms, which is MUCH faster compared with 0.932 ms for the current 64K grouped head.
> 
> Results varied though with acceptance because my proposal trace is probably still too small, so I'm looking to do broader testing before jumping into the converter and model format.

I guess you can adjust tokenizer mix in a small 16K 32K based on user‘s usage data. you collect the data during your daily usage like very long agentic coding session （There are plenty of repetition） and applied it accordingly.

PR #306, opened 2026-08-24
Fix Unigram tokenizer unknown token handling

### LINKED ISSUE #305: Unigram tokenizers fail because tokenizer.model.unk_token is assumed to exist
## Summary

ExLlamaV3 assumes that the underlying Hugging Face tokenizer model exposes an `unk_token` attribute:

```python
ut = self.tokenizer.model.unk_token
```

However, `tokenizers.models.Unigram` does not expose `unk_token`.

This causes tokenizer initialization to fail for models that use a Hugging Face Unigram tokenizer.

I encountered this while trying to convert `llm-jp/llm-jp-4-33b-base`, a Llama-based causal language model, to EXL3. Its tokenizer is based on the Hugging Face Tokenizers Unigram byte-fallback model.

## Minimal reproduction of the underlying issue

The behavior of `tokenizers.models.Unigram` itself can be demonstrated with a standard Unigram tokenizer such as T5:

```python
from tokenizers import Tokenizer

tokenizer = Tokenizer.from_pretrained("google-t5/t5-base")

print(type(tokenizer.model))
# => <class 'tokenizers.models.Unigram'>
print(hasattr(tokenizer.model, "unk_token"))
# => False
print(tokenizer.model.unk_token)
# => AttributeError: 'tokenizers.models.Unigram' object has no attribute 'unk_token'
```

The model is a `tokenizers.models.Unigram`, and accessing `unk_token` raises:

```text
AttributeError: 'tokenizers.models.Unigram' object has no attribute 'unk_token'
```

T5 itself is not the ExLlamaV3 conversion use case here; it is only a small example demonstrating the underlying Hugging Face Tokenizers API behavior.

## Real-world impact

I encountered the issue while converting:

[https://huggingface.co/llm-jp/llm-jp-4-33b-base](https://huggingface.co/llm-jp/llm-jp-4-33b-base)

Using the normal conversion command:

```bash
python convert.py \
    -i /path/to/llm-jp-4-33b-base \
    -o /path/to/output \
    -w /path/to/work \
    -b 4.0
```

conversion fails during tokenizer initialization, before quantization can proceed.

The resulting error is:

```text
AttributeError: 'tokenizers.models.Unigram' object has no attribute 'unk_token'
```

The model card describes its tokenizer as being based on the Hugging Face Tokenizers Unigram byte-fallback model.

## Expected behavior

ExLlamaV3 should be able to initialize Hugging Face Unigram tokenizers without assuming that every tokenizer model implementation exposes `unk_token`.

## Actual behavior

Tokenizer initialization directly accesses:

```python
self.tokenizer.model.unk_token
```

which is not available on `tokenizers.models.Unigram`, causing initialization to fail.

## Possible fix

The unknown-token handling may need to account for differences between tokenizer model implementations rather than directly assuming the presence of `model.unk_token`.

For example, the Unigram API uses unk_id rather than unk_token, so this may require model-specific handling.

I am currently testing a fix and can submit a PR once I have verified it against the affected model and existing tokenizer types.

## Environment

- ExLlamaV3 commit: `2398c05635fbbad01a0a51dce63c85c6c8a8450e`
- Python: `3.12.10`
- tokenizers: `0.22.2`

shunmoridev:
Fixes #305

Hugging Face `Unigram` tokenizer models do not expose `model.unk_token`.
For Unigram tokenizers, resolve the unknown token using the serialized
`model.unk_id` stored in `tokenizer.json` instead.

The existing behavior for other tokenizer model types is unchanged.

The existing `m` variable, which is already assigned to
`self.tokenizer.model` earlier in the initializer, is reused here as well.

Testing:

- Reproduced the original `AttributeError` with `llm-jp/llm-jp-4-33b-base` before the fix.
- Verified that its Unigram tokenizer initializes successfully with the fix.
- Verified that conversion proceeds past tokenizer initialization.
- Verified that `Qwen/Qwen3-0.6B-Base`, using the existing non-Unigram tokenizer path, converts successfully.

### DIFF
--- a/exllamav3/tokenizer/tokenizer.py
+++ b/exllamav3/tokenizer/tokenizer.py
@@ -112,8 +112,15 @@ class Tokenizer:
         self.missing_special_delimiters = None
 
         # Get control token IDs
-        ut = self.tokenizer.model.unk_token
-        self.unk_token_id = None if ut is None else self.tokenizer.token_to_id(ut)
+        if isinstance(m, models.Unigram):
+            # Unigram stores the unknown token as an ID in tokenizer.json
+            model_config = maybe_read_json(self.path_tokenizer_json).get("model", {})
+            unk_id = model_config.get("unk_id")
+            ut = None if unk_id is None else self.tokenizer.id_to_token(unk_id)
+        else:
+            ut = m.unk_token
+            unk_id = None if ut is None else self.tokenizer.token_to_id(ut)
+        self.unk_token_id = unk_id
         self.eos_token_id = self.config.eos_token_id
         self.bos_token_id = self.config.bos_token_id
         self.pad_token_id = self.config.pad_token_id
@@ -158,7 +165,7 @@ class Tokenizer:
                     config.eos_token_id_list.append(e)
 
         # Get control token strings
-        self.unk_token = self.tokenizer.model.unk_token
+        self.unk_token = ut
         self.bos_token = None if self.bos_token_id is None else \
             (self.extended_id_to_piece.get(self.bos_token_id) or self.tokenizer.id_to_token(self.bos_token_id))
         self.eos_token = None if self.eos_token_id is None else \


turboderp:
Thanks! :+1:

PR #311, opened 2026-08-31
Precomputed MoE expert-placement profiles for CPU offload

anoane:
Adds optional precomputed expert-placement profiles for CPU MoE offload. Additive and inert
unless `--moe_cpu_profile` is passed; no behaviour changes on any existing path.

## Problem

With `-mcs`, which experts stay resident decides how much traffic crosses the memory bus, and
on a host-DRAM-bound setup that is most of decode. Placement is currently discovered at
runtime, and convergence is slow by construction: `_split_swap_tick` moves at most
`EXL3_MOE_CPU_SWAP_MAX` (64) experts across **all** layers every `EXL3_MOE_CPU_SWAP_INTERVAL`
(128) decode steps, and each promotion re-reads the expert from the checkpoint on disk. On a
42-layer model that is ~1.5 experts per layer per sweep, so a long generation can end before
placement settles — and a fresh request starts over.

## What this adds

A profile is a per-expert selection census measured offline and loaded at startup, so
placement is correct at token zero. Nothing is probed or computed at load beyond one lookup
per layer.

```
-mcp  / --moe_cpu_profile            comma-separated profiles, optional :weight each
-mcpm / --moe_cpu_profile_mode       static (freeze the order) | seed (start hot, keep adapting)
-mcpd / --moe_cpu_profile_dir        extra search path
-mcpq / --moe_cpu_profile_any_quant  allow a checkpoint mismatch, with a warning
```

Profiles resolve from `<model_dir>/moe_profiles`, `$EXL3_MOE_PROFILE_DIR`, or a path.
`seed` mode is new capability: a profile is currently refused unless swapping is off.

Diff to existing files is +107/-10 across `model_init.py` and `block_sparse_mlp_cpu.py`.
Everything else is new: the loader, an offline builder, a packer, and tests.

## Measured

GLM-5.3-Flash-exl3 4.05bpw (45 layers, 42 MoE, 288 experts, topk 8) on one RTX PRO 6000
Blackwell 96GB + Ryzen 9 9950X3D / 128GB DDR5. Driven through the Generator on held-out text,
`-cs 270336`, warm, first generation discarded.

| workload | context | `-mcs 180`, no profile | `-mcs 128` + profile | |
|---|---:|---:|---:|---:|
| prose | 32,768 | 17.01 tok/s | 32.91 | 1.93x |
| prose | 262,144 | 16.82 / 16.95 | 40.24 / 39.71 | 2.37x |
| code | 32,768 | 16.82 | 32.78 | 1.95x |
| code | 262,144 | 17.11 / 17.42 | 33.66 / 34.01 | 1.96x |

Prefill improves as well, since resident experts cut CPU work during prefill: 573 -> 771
tok/s (prose 32k), 696 -> 889 (prose 256k).

About 1.35x of that is residency (`-mcs 180` -> `128`), which needs no code — 160 resident of
288 is simply the most that loads at `-cs 270336` on this card. The remaining ~1.64x is
placement, which is what this PR provides.

## Why it works

Decode time is a straight line in cold-expert rate. Fitted over 14 points and cross-validated
on 16 more, R² = 0.9994:

```
step_ms  ≈  10.8  +  0.77 × cold_percent        cold_percent = (cpu-assign/row)/8 × 100
```

The slope is exactly `42 layers × 8 topk × 12.62 MB / 100 / 53 GB/s` — the fit recovers the
measured AVX-512 CPU kernel bandwidth on its own, which is a decent check that the model is
describing the real mechanism rather than curve-fitting.

Nothing else in the system moves with context. Per-module CUDA-event timing from 4,096 to
261,888 tokens: attention/DSA **+0.137 ms**, KDA +0.010, hyper-connections +0.003, norms
+0.007, lm_head 0.000. At 262,144 the step is ~38 ms of exposed CPU-expert stall against
~14.5 ms of GPU work, of which attention is 0.95 ms. Cold rate is the only term that matters,
so anything that lowers it converts directly into throughput.

## Scoring is held-out by construction

This matters more than it might look. Fitting a profile to one prompt and reporting how well
that prompt's own hot set covers that same prompt's routing measures nothing — the hot set is
chosen to cover it, so the number reads ~93% regardless of whether the ranking generalizes.
Measured live, a profile built that way captured 42-45% against 37.5% for random placement,
and was **slower than using no profile at all**: it pinned a bad hot set and suppressed the
existing sweeps from finding a better one.

`util/moe_profile_build.py` therefore samples many independent windows, ranks on a subset, and
reports capture on the disjoint remainder, beside two reference points:

- `uniform` = `R/E`, what random placement scores. A profile at this level is worthless.
- `oracle` = fit on the held-out set itself: the ceiling for any static profile.

It warns when held-out sits within 3 points of uniform, or when in-sample exceeds held-out by
more than 25 points, and refuses to report capture below 4 windows rather than printing an
in-sample number. On the reference profile, held-out predicted 32.9% cold and the live
measurement was 31.7%.

## A profile must match the workload on two axes

Both measured, and getting either wrong costs the entire benefit:

**Domain.** A prose census applied to code is worth 1.01x — nothing. A code census on code is
1.49x.

**Context length.** A census fitted on 1k windows loses most of its value by 256k. This is a
length effect, not a content artifact: holding the trailing 4096 tokens byte-identical and
varying only the preceding context still moved cold rate 20.8% -> 28.6%.

There is a further result worth stating, since it is counterintuitive: profiles must be built
per **register**, not per topic. Six profiles built by merging related corpora each scored
below their own members, ordered by how mixed their registers were. A nine-category "security"
profile landed at 14% of available headroom — barely above random — while eight of its nine
constituent corpora reach 55-80% individually. A seven-category "systems" profile kept 60%,
because those share one language family.

## Format

`.exl3moe` — safetensors carrying the precomputed ranking, the summed counts, and the
per-prompt census, each in its own byte range. Serving preads only `ranking`; `-score` preads
only `census` and re-runs the held-out split with no model load. `build_ranking()` end to end
is 0.0534 ms against 3.1892 ms for the equivalent `.npz`, which has to be decompressed in
full, summed over prompts and sorted every time. Rankings are bit-identical. Resolution
prefers `.exl3moe` -> `.safetensors` -> `.npz` -> `.json`, and `.npz` is still accepted for
interop with external tooling.

The reader uses `pread` rather than mmap and names the tensors it wants: at 147 KB (36 pages)
mmap pays a VMA setup plus a fault per page where pread is one syscall and a copy — 0.0080 vs
0.0181 ms — and `safetensors.load_file` additionally opens the file a second time for
metadata. That matches what `exllamav3_ext/stloader.cpp` already does for bulk weights.

## Identity

Profiles are per `(model, checkpoint)`. Model identity (architecture, layers, experts,
`moe_intermediate_size`, `hidden_size`) is fatal on mismatch even with the override.
Checkpoint identity (`checkpoint_sha`, `quant_method`, `bits`, `head_bits`, `codebook`) is
fatal unless `--moe_cpu_profile_any_quant`. `checkpoint_sha` hashes safetensors **headers**
only — 0.03 s and 19.7 MB of reads on a 165 GB checkpoint, not a content hash.

## Tests

`tests/test_moe_profile.py`, 40 cases, no CUDA or built extension required. The load-bearing
one is `test_capture_exposes_per_prompt_overfit`: it constructs the exact pathology described
above — each prompt hot on its own experts — and asserts held-out capture collapses to chance
while in-sample still looks strong.

## Note on measuring this

`eval/perf.py` is not usable for placement or long-context claims: `measure_generate` builds
state via `cache.get_test_state()`, which passes `clear=True`, so 34 of 45 recurrent layers
decode from zero state and its long-context routing is really short-context routing (cold
29.5% at 261,888 against ~55% from the Generator on the same setup). Drive the Generator
instead, and report `cpu-assign/row` (`EXL3_MOE_HANDOFF_PROF=1`) next to any throughput
number — tok/s alone cannot separate a placement effect from a warm-up artifact.

## Worked example

A full serving recipe with 27 profiles for GLM-5.3-Flash-exl3 4.05bpw, one per register, each
with its held-out capture number, plus the measurement harness and documentation on building
and loading them: https://github.com/anoane/glm53-flash-exl3-recipe

## Limitations

- Profiles are model- and checkpoint-specific, so this ships tooling rather than data.
- A profile built for the wrong domain or context length is worth roughly nothing; the
  scoring is designed to make that visible before anything is deployed, but it cannot be
  detected automatically at load.
- `seed` mode is exposed but measured no better than `static` on this model: 24.3% vs 24.9%
  cold across 8 turns at fixed context, inside noise. Sweeping mid-stream also perturbs
  generation (perplexity 3.4924 vs 3.4904 static), which is consistent with the existing
  comment about deferring sweeps to queue-drain.
- Only tested on one model and one machine. The mechanism is general — any host-DRAM-bound
  offload setup — but the constants in the cost model are not.

### DIFF
--- a/doc/moe_expert_profiles.md
+++ b/doc/moe_expert_profiles.md
@@ -1,0 +1,189 @@
+# MoE expert placement profiles
+
+CPU MoE offload (`-mcs` / `-mcl`) splits each block-sparse layer's experts between VRAM and
+system RAM. Routing is heavily skewed, so *which* experts stay resident decides how much
+traffic crosses the memory bus — on a DDR5-bound setup that is the difference between the CPU
+being invisible and the CPU being the bottleneck.
+
+Upstream discovers that ordering at runtime. A **profile** measures it ahead of time, so
+placement is right at token zero.
+
+## Why placement matters
+
+Measured on GLM-5.3-Flash (42 MoE layers × 288 experts), sorting each layer's experts by
+measured decode traffic:
+
+| experts kept in VRAM | `-mcs` | traffic served from VRAM | CPU share | vs uniform placement |
+| --- | --- | --- | --- | --- |
+| 75% | 72 | 98.4% | 1.6% | 15.2x less CPU work |
+| 50% | 144 | 90.7% | 9.3% | 5.4x |
+| 37.5% | 180 | 83.7% | 16.3% | 3.8x |
+| 25% | 216 | 73.5% | 26.5% | 2.8x |
+
+The curve is strongly non-linear: half the experts can be evicted while ~91% of the routing
+still lands in VRAM.
+
+## Measured effect
+
+Placement only buys throughput where the CPU side is actually the bottleneck. On GLM-5.3-Flash,
+`eval/perf.py`, decode tok/s averaged over context 0..4k, native profile, two repeats:
+
+GLM-5.3-Flash 4.05bpw, `-mcs 180`, native profile, decode tok/s:
+
+| context | dynamic | `seed` + profile | |
+| --- | --- | --- | --- |
+| 0 | 18.80 | 21.88 | +16.4% |
+| 256 | 18.52 | 20.79 | +12.3% |
+| 512 | 18.82 | 21.00 | +11.6% |
+| 1,024 | 18.52 | 19.82 | +7.0% |
+| 2,048 | 15.37 | 17.51 | +13.9% |
+| 4,096 | 17.19 | 16.76 | -2.5% |
+| 8,192 | 18.29 | 20.02 | +9.5% |
+| 16,384 | 17.00 | 18.72 | +10.1% |
+| 32,768 | 17.05 | 17.95 | +5.3% |
+| 65,536 | 15.26 | 14.73 | -3.5% |
+| 131,072 | 18.09 | 18.91 | +4.5% |
+| 261,888 | 17.40 | 17.69 | +1.7% |
+
+| band | dynamic | `seed` | | |
+| --- | --- | --- | --- | --- |
+| short, 0-2k | 18.01 | 20.20 | **+12.2%** | wins 5/5 |
+| mid, 4k-32k | 17.38 | 18.36 | +5.6% | wins 3/4 |
+| long, 64k-256k | 16.92 | 17.11 | +1.1% | wins 2/3 |
+| overall | 17.44 | 18.73 | +7.4% | wins 10/12 |
+
+**The gain decays with context, and that is structural.** Placement only affects the MoE share
+of a decode step. That work is constant per token while attention grows with context, so by
+64k the experts are a small part of the critical path and reordering them buys little. Build
+profiles for chat, agentic loops and code completion; skip them for long-document work.
+
+The same run on 2.05bpw `-mcs 150` gives +3.5% (5/7 points, not significant): that config has
+DDR5 bandwidth to spare, so there is nothing to reclaim. The 4.05bpw config decodes at ~95% of
+its bandwidth ceiling, which is why it responds. Measure your own configuration.
+
+Correctness is unaffected: perplexity over the same rows is 3.489850 with no offload, 3.490881
+with offload and dynamic placement, 3.492384 seeded and 3.490442 static -- all within eval
+noise.
+
+## Building a profile
+
+```
+python util/moe_profile_build.py -m /models/GLM-5.3-Flash-exl3-4.05bpw \
+    -corpus /data/wikitext-2-raw/wiki.train.raw -o wiki.npz \
+    -nprompts 48 -plen 1024 -gen 128 -resident 108
+```
+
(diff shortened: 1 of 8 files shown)


anoane:
Closing in favour of a rebuilt branch. Serving a model with a packed profile surfaced a load-time failure this PR contained: safetensors metadata is string-valued, so the packer stored `fingerprint` as JSON text and the loader passed that string to `check_fingerprint()`, which indexes it by key — `TypeError: string indices must be integers`. That fired for every `.exl3moe` profile and defeated the checkpoint-identity check the format exists to carry. `.npz` profiles were unaffected, which is why the packing and ranking tests missed it: none exercised identity validation on the packed path.

Fixed, and every documented flag path is now exercised against real profile files rather than synthetic ones — name/path/`$EXL3_MOE_PROFILE_DIR` resolution, single and weighted merges including order-independence and weight-relativity, fingerprint policy under both strictness settings, and that every shipped ranking is a valid permutation. Test count 40 → 41.

Re-opening as a single clean commit rather than stacking a fixup on a broken base.

PR #312, opened 2026-08-31
Adds pyproject.toml packaging

ottergrad:


ottergrad:
[Here's the manual build](https://github.com/ottergrad/exllamav3/actions/runs/33445128013) I triggered with the updated workflow, both Linux and Windows build targets succeeded.

turboderp:
Some issues with `uv sync` it looks like. The lockfile would pin torch 2.13.0 from PyPi (replacing the specific version the readme says to install just prior.) There's really no way to meaningfully pin a Torch version, I believe. (?) I think this would also break Windows installs since the wheel it pulls there would be CPU only. And I'm not sure about having to maintain the lockfile with each new dependency update. (?)

Sorry, packaging breaks my smooth brain. 8| But I'm leaning towards omitting uv.lock and recommending `uv pip install .`

I think JIT mode would also be broken because `exllamav3_ext/*` isn't included. And I believe you might also need `py.typed` and `util/hadamard_data/` though I'm less sure if it's actually used anywhere.

The python>=3.11 requirement is okay, but more precisely it could require >=3.10.11, since that's the minimum version that should work with the recent changes to flash-linear-attention.

And also on Windows you'd want to pull `triton-windows` since pyproject at least allows for that.

ottergrad:
I actually did consider omitting the lock file. Let me think about that more but you're probably right to just drop it.

JIT mode works with an editable install, but in a non-editable install from source maybe not? I can test that, if that's what you're talking about.

I make the other changes.

ottergrad:
Other than the `py.typed` and`util/hadamard_data/`, I think this addresses your comments? I'll do some more manual testing to validate the install commands.

What do you think?

ottergrad:
Latest wheel builds (test, cu132 only)
https://github.com/ottergrad/exllamav3/actions/runs/33787731238

Latest sdist build
https://github.com/ottergrad/exllamav3/actions/runs/33794446548

ottergrad:
I tested all but the `pip` based install commands (I don't even have it on my machine, `uv` enjoyer) and they work great. I tested by installing using each command in a fresh environment and loading up a small model for inference, triggering JIT, and all paths went fine (no errors, model inferenced correctly).

PR #315, opened 2026-09-01
Precomputed MoE expert-placement profiles for CPU offload

anoane:
Adds optional precomputed expert-placement profiles for CPU MoE offload. Additive and inert
unless `--moe_cpu_profile` is passed; no behaviour changes on any existing path.

## Problem

With `-mcs`, which experts stay resident decides how much traffic crosses the memory bus, and
on a host-DRAM-bound setup that is most of decode. Placement is currently discovered at
runtime, and convergence is slow by construction: `_split_swap_tick` moves at most
`EXL3_MOE_CPU_SWAP_MAX` (64) experts across **all** layers every `EXL3_MOE_CPU_SWAP_INTERVAL`
(128) decode steps, and each promotion re-reads the expert from the checkpoint on disk. On a
42-layer model that is ~1.5 experts per layer per sweep, so a long generation can end before
placement settles — and a fresh request starts over.

## What this adds

A profile is a per-expert selection census measured offline and loaded at startup, so
placement is correct at token zero. Nothing is probed or computed at load beyond one lookup
per layer.

```
-mcp  / --moe_cpu_profile            comma-separated profiles, optional :weight each
-mcpm / --moe_cpu_profile_mode       static (freeze the order) | seed (start hot, keep adapting)
-mcpd / --moe_cpu_profile_dir        extra search path
-mcpq / --moe_cpu_profile_any_quant  allow a checkpoint mismatch, with a warning
```

Profiles resolve from `<model_dir>/moe_profiles`, `$EXL3_MOE_PROFILE_DIR`, or a path.
`seed` mode is new capability: a profile is currently refused unless swapping is off.

Diff to existing files is +107/-10 across `model_init.py` and `block_sparse_mlp_cpu.py`.
Everything else is new: the loader, an offline builder, a packer, and tests.

## Measured

GLM-5.3-Flash-exl3 4.05bpw (45 layers, 42 MoE, 288 experts, topk 8) on one RTX PRO 6000
Blackwell 96GB + Ryzen 9 9950X3D / 128GB DDR5. Driven through the Generator on held-out text,
`-cs 270336`, warm, first generation discarded.

| workload | context | `-mcs 180`, no profile | `-mcs 128` + profile | |
|---|---:|---:|---:|---:|
| prose | 32,768 | 17.01 tok/s | 32.91 | 1.93x |
| prose | 262,144 | 16.82 / 16.95 | 40.24 / 39.71 | 2.37x |
| code | 32,768 | 16.82 | 32.78 | 1.95x |
| code | 262,144 | 17.11 / 17.42 | 33.66 / 34.01 | 1.96x |

Prefill improves as well, since resident experts cut CPU work during prefill: 573 -> 771
tok/s (prose 32k), 696 -> 889 (prose 256k).

About 1.35x of that is residency (`-mcs 180` -> `128`), which needs no code — 160 resident of
288 is simply the most that loads at `-cs 270336` on this card. The remaining ~1.64x is
placement, which is what this PR provides.

## Why it works

Decode time is a straight line in cold-expert rate. Fitted over 14 points and cross-validated
on 16 more, R² = 0.9994:

```
step_ms  ≈  10.8  +  0.77 × cold_percent        cold_percent = (cpu-assign/row)/8 × 100
```

The slope is exactly `42 layers × 8 topk × 12.62 MB / 100 / 53 GB/s` — the fit recovers the
measured AVX-512 CPU kernel bandwidth on its own, which is a decent check that the model is
describing the real mechanism rather than curve-fitting.

Nothing else in the system moves with context. Per-module CUDA-event timing from 4,096 to
261,888 tokens: attention/DSA **+0.137 ms**, KDA +0.010, hyper-connections +0.003, norms
+0.007, lm_head 0.000. At 262,144 the step is ~38 ms of exposed CPU-expert stall against
~14.5 ms of GPU work, of which attention is 0.95 ms. Cold rate is the only term that matters,
so anything that lowers it converts directly into throughput.

## Scoring is held-out by construction

This matters more than it might look. Fitting a profile to one prompt and reporting how well
that prompt's own hot set covers that same prompt's routing measures nothing — the hot set is
chosen to cover it, so the number reads ~93% regardless of whether the ranking generalizes.
Measured live, a profile built that way captured 42-45% against 37.5% for random placement,
and was **slower than using no profile at all**: it pinned a bad hot set and suppressed the
existing sweeps from finding a better one.

`util/moe_profile_build.py` therefore samples many independent windows, ranks on a subset, and
reports capture on the disjoint remainder, beside two reference points:

- `uniform` = `R/E`, what random placement scores. A profile at this level is worthless.
- `oracle` = fit on the held-out set itself: the ceiling for any static profile.

It warns when held-out sits within 3 points of uniform, or when in-sample exceeds held-out by
more than 25 points, and refuses to report capture below 4 windows rather than printing an
in-sample number. On the reference profile, held-out predicted 32.9% cold and the live
measurement was 31.7%.

## A profile must match the workload on two axes

Both measured, and getting either wrong costs the entire benefit:

**Domain.** A prose census applied to code is worth 1.01x — nothing. A code census on code is
1.49x.

**Context length.** A census fitted on 1k windows loses most of its value by 256k. This is a
length effect, not a content artifact: holding the trailing 4096 tokens byte-identical and
varying only the preceding context still moved cold rate 20.8% -> 28.6%.

There is a further result worth stating, since it is counterintuitive: profiles must be built
per **register**, not per topic. Six profiles built by merging related corpora each scored
below their own members, ordered by how mixed their registers were. A nine-category "security"
profile landed at 14% of available headroom — barely above random — while eight of its nine
constituent corpora reach 55-80% individually. A seven-category "systems" profile kept 60%,
because those share one language family.

## Format

`.exl3moe` — safetensors carrying the precomputed ranking, the summed counts, and the
per-prompt census, each in its own byte range. Serving preads only `ranking`; `-score` preads
only `census` and re-runs the held-out split with no model load. `build_ranking()` end to end
is 0.0534 ms against 3.1892 ms for the equivalent `.npz`, which has to be decompressed in
full, summed over prompts and sorted every time. Rankings are bit-identical. Resolution
prefers `.exl3moe` -> `.safetensors` -> `.npz` -> `.json`, and `.npz` is still accepted for
interop with external tooling.

The reader uses `pread` rather than mmap and names the tensors it wants: at 147 KB (36 pages)
mmap pays a VMA setup plus a fault per page where pread is one syscall and a copy — 0.0080 vs
0.0181 ms — and `safetensors.load_file` additionally opens the file a second time for
metadata. That matches what `exllamav3_ext/stloader.cpp` already does for bulk weights.

Structured metadata (`fingerprint`, `capture`) is stored as JSON text, since safetensors
metadata is string-valued, and is decoded on load. Returning it unparsed made
`check_fingerprint` index a `str` and raise `string indices must be integers` at model load
for every packed profile — which also defeated the checkpoint-identity check the format exists
to carry. `.npz` profiles were unaffected, their fingerprint arriving already decoded from the
sidecar.

## Identity

Profiles are per `(model, checkpoint)`. Model identity (architecture, layers, experts,
`moe_intermediate_size`, `hidden_size`) is fatal on mismatch even with the override.
Checkpoint identity (`checkpoint_sha`, `quant_method`, `bits`, `head_bits`, `codebook`) is
fatal unless `--moe_cpu_profile_any_quant`. `checkpoint_sha` hashes safetensors **headers**
only — 0.03 s and 19.7 MB of reads on a 165 GB checkpoint, not a content hash.

## Tests

`tests/test_moe_profile.py`, 41 cases, no CUDA or built extension required. The load-bearing
one is `test_capture_exposes_per_prompt_overfit`: it constructs the exact pathology described
above — each prompt hot on its own experts — and asserts held-out capture collapses to chance
while in-sample still looks strong.

Every documented flag path was additionally exercised against real profile files rather than
synthetic ones, which is how the metadata bug above was found — by serving a model, not by
testing. Covered: name / absolute path / `$EXL3_MOE_PROFILE_DIR` resolution and that an
external-only profile stays invisible without it; single-source and weighted merges, including
that a merge actually differs from its inputs, is order-independent at equal weights, and
treats weights as relative (`3:1` == `6:2`); fingerprint policy under both strictness settings,
including that model-identity mismatch stays fatal under `--moe_cpu_profile_any_quant`; and
that every shipped ranking is a valid permutation.

## Note on measuring this

`eval/perf.py` is not usable for placement or long-context claims: `measure_generate` builds
state via `cache.get_test_state()`, which passes `clear=True`, so 34 of 45 recurrent layers
decode from zero state and its long-context routing is really short-context routing (cold
29.5% at 261,888 against ~55% from the Generator on the same setup). Drive the Generator
instead, and report `cpu-assign/row` (`EXL3_MOE_HANDOFF_PROF=1`) next to any throughput
number — tok/s alone cannot separate a placement effect from a warm-up artifact.

## Worked example

A full serving recipe with 27 profiles for GLM-5.3-Flash-exl3 4.05bpw, one per register, each
with its held-out capture number, plus the measurement harness and documentation on building
and loading them: https://github.com/anoane/glm53-flash-exl3-recipe

## Limitations

- Profiles are model- and checkpoint-specific, so this ships tooling rather than data.
- A profile built for the wrong domain or context length is worth roughly nothing; the
  scoring is designed to make that visible before anything is deployed, but it cannot be
  detected automatically at load.
- `seed` mode is exposed but measured no better than `static` on this model: 24.3% vs 24.9%
  cold across 8 turns at fixed context, inside noise. Sweeping mid-stream also perturbs
  generation (perplexity 3.4924 vs 3.4904 static), which is consistent with the existing
  comment about deferring sweeps to queue-drain.
- Only tested on one model and one machine. The mechanism is general — any host-DRAM-bound
  offload setup — but the constants in the cost model are not.

### DIFF
--- a/doc/moe_expert_profiles.md
+++ b/doc/moe_expert_profiles.md
@@ -1,0 +1,189 @@
+# MoE expert placement profiles
+
+CPU MoE offload (`-mcs` / `-mcl`) splits each block-sparse layer's experts between VRAM and
+system RAM. Routing is heavily skewed, so *which* experts stay resident decides how much
+traffic crosses the memory bus — on a DDR5-bound setup that is the difference between the CPU
+being invisible and the CPU being the bottleneck.
+
+Upstream discovers that ordering at runtime. A **profile** measures it ahead of time, so
+placement is right at token zero.
+
+## Why placement matters
+
+Measured on GLM-5.3-Flash (42 MoE layers × 288 experts), sorting each layer's experts by
+measured decode traffic:
+
+| experts kept in VRAM | `-mcs` | traffic served from VRAM | CPU share | vs uniform placement |
+| --- | --- | --- | --- | --- |
+| 75% | 72 | 98.4% | 1.6% | 15.2x less CPU work |
+| 50% | 144 | 90.7% | 9.3% | 5.4x |
+| 37.5% | 180 | 83.7% | 16.3% | 3.8x |
+| 25% | 216 | 73.5% | 26.5% | 2.8x |
+
+The curve is strongly non-linear: half the experts can be evicted while ~91% of the routing
+still lands in VRAM.
+
+## Measured effect
+
+Placement only buys throughput where the CPU side is actually the bottleneck. On GLM-5.3-Flash,
+`eval/perf.py`, decode tok/s averaged over context 0..4k, native profile, two repeats:
+
+GLM-5.3-Flash 4.05bpw, `-mcs 180`, native profile, decode tok/s:
+
+| context | dynamic | `seed` + profile | |
+| --- | --- | --- | --- |
+| 0 | 18.80 | 21.88 | +16.4% |
+| 256 | 18.52 | 20.79 | +12.3% |
+| 512 | 18.82 | 21.00 | +11.6% |
+| 1,024 | 18.52 | 19.82 | +7.0% |
+| 2,048 | 15.37 | 17.51 | +13.9% |
+| 4,096 | 17.19 | 16.76 | -2.5% |
+| 8,192 | 18.29 | 20.02 | +9.5% |
+| 16,384 | 17.00 | 18.72 | +10.1% |
+| 32,768 | 17.05 | 17.95 | +5.3% |
+| 65,536 | 15.26 | 14.73 | -3.5% |
+| 131,072 | 18.09 | 18.91 | +4.5% |
+| 261,888 | 17.40 | 17.69 | +1.7% |
+
+| band | dynamic | `seed` | | |
+| --- | --- | --- | --- | --- |
+| short, 0-2k | 18.01 | 20.20 | **+12.2%** | wins 5/5 |
+| mid, 4k-32k | 17.38 | 18.36 | +5.6% | wins 3/4 |
+| long, 64k-256k | 16.92 | 17.11 | +1.1% | wins 2/3 |
+| overall | 17.44 | 18.73 | +7.4% | wins 10/12 |
+
+**The gain decays with context, and that is structural.** Placement only affects the MoE share
+of a decode step. That work is constant per token while attention grows with context, so by
+64k the experts are a small part of the critical path and reordering them buys little. Build
+profiles for chat, agentic loops and code completion; skip them for long-document work.
+
+The same run on 2.05bpw `-mcs 150` gives +3.5% (5/7 points, not significant): that config has
+DDR5 bandwidth to spare, so there is nothing to reclaim. The 4.05bpw config decodes at ~95% of
+its bandwidth ceiling, which is why it responds. Measure your own configuration.
+
+Correctness is unaffected: perplexity over the same rows is 3.489850 with no offload, 3.490881
+with offload and dynamic placement, 3.492384 seeded and 3.490442 static -- all within eval
+noise.
+
+## Building a profile
+
+```
+python util/moe_profile_build.py -m /models/GLM-5.3-Flash-exl3-4.05bpw \
+    -corpus /data/wikitext-2-raw/wiki.train.raw -o wiki.npz \
+    -nprompts 48 -plen 1024 -gen 128 -resident 108
+```
+
+`-corpus` takes a bundled calibration corpus (`wiki`, `c4`, `code`, `multilingual`,
+`technical`, `tiny`), a text file, or a JSON list of prompts. The bundled `.utf8` files are not
+installed as package data, so on an installed exllamav3 pass a path or set `EXL3_CAL_DATA_DIR`.
+A model that does not fit resident needs `-mcs`; routing is unaffected by where experts live:
+only routing decisions are needed and a fully resident load is faster.
+
+Decode is the allocation signal — that is the regime offload runs in. Profile quality is driven
+first by **how many independent windows** are sampled (`-nprompts`) and only then by how many
+tokens each contributes (`-plen`, `-gen`). Diversity beats length: one long generation follows a
+single trajectory and its ranking encodes that trajectory rather than the corpus. A profile fitted
+that way scored 94.8% against its own counts and delivered 42-45% live — worse than random
+placement plus upstream's own sweeps.
+
+Two guards make that failure visible instead of silent:
+
+* **Sample size.** Below 20 hits/expert the tool warns and falls back to the prefill bank, because
+  a ranking built from one or two hits per expert is noise that *looks* like extreme concentration
+  (unsampled experts sort to the tail, so the "hot" head trivially covers everything).
+* **Generalization.** Windows are split into disjoint fit/test sets; capture is reported on the
+  held-out half beside `uniform` (random placement) and `oracle` (the best any static profile can
+  do). Held-out within 3 points of uniform, or in-sample more than 25 points above held-out, each
+  print an explicit warning. With fewer than 4 windows no split is possible and the tool says so
+  rather than reporting an in-sample number.
+
+## Flags
+
+| flag | meaning |
+| --- | --- |
+| `-mcp` / `--moe_cpu_profile` | profile spec: names or paths, comma-separated, optional `:weight` |
+| `-mcpm` / `--moe_cpu_profile_mode` | `seed` (default) or `static` |
+| `-mcpd` / `--moe_cpu_profile_dir` | extra search directory |
+| `-mcpq` / `--moe_cpu_profile_any_quant` | accept a profile from another checkpoint of the same model |
+
+## Using a profile
+
+```
+python -m exllamav3.server -m /models/... -mcs 180 \
+    --moe_cpu_profile code --moe_cpu_profile_mode seed
+```
+
+Profiles are looked up as `<name>.{npz,safetensors,exl3moe,json}`, in order:
+
+1. `<model_dir>/moe_profiles/` — ships inside the model repo
+2. `$EXL3_MOE_PROFILE_DIR` (os.pathsep-separated)
+3. `~/.cache/exllamav3/moe_profiles/`
+
+### Combining profiles
+
+```
+--moe_cpu_profile code:3,wiki:1
+```
+
+Sources are merged on **per-layer normalized** frequency, so a corpus with more tokens does not
+simply win; the optional `:weight` sets relative contribution.
+
+### seed vs static
+
+| mode | placement | dynamic swapping | use when |
+| --- | --- | --- | --- |
+| `-mcpm seed` (default) | starts from the profile | stays on, keeps adapting | the workload may drift from the profile |
+| `-mcpm static` | frozen at the profile order | off | you want predictable latency and no mid-generation placement changes |
+
+`static` is the more stable of the two: a placement change is a small numeric step for every
+migrated expert (the same weights compute slightly differently on the two devices), so `seed`
+can shift output subtly at a sweep boundary while `static` never does.
+
+## Profile identity: per model, per quantization
+
+Routing depends on the model **and** on how it was quantized — different bitrates give
+different hidden states and therefore different expert choices. A profile therefore belongs to
+one `(model, checkpoint)` pair, and both are checked at load:
+
+* **Model identity** — architecture, layer count, expert count, `moe_intermediate_size`,
+  `hidden_size`. A mismatch is always fatal; placement from another architecture is meaningless.
+* **Checkpoint identity** — `checkpoint_sha` plus `quant_method`/`bits`/`head_bits`/`codebook`.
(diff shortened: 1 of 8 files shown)


joaosump:
Not sure why this is not merged yet, I think it's a great granular control over the layers that are prioritised for VRAM, works quite well for me and Qwen 3.8 Flash Next

turboderp:
It's mostly not merged because it's taking a while for me to get to it. It adds 1700 lines of text/code to review and adds what seems like a lot of new surface area and tech debt. It does seem useful, and I want to include it, but I need to decide whether this is a secret feature buried deep in some .md file that 1% of users will notice or if it gets streamlined and exposed to the other 99% through a simple (somehow) interface. Maybe it's just a default-on, opt-out feature? Maybe expert heatmaps are computed automatically during quantization? Or maybe it could be made redundant by just forcing more expert swaps early on after a model is loaded?

I do hope to answer these questions soon and then make some progress. But it all takes time.

joaosump:
@turboderp my comment might have come out as a sort of a complaining which wasn't my intention at all. I cannot imagine the amount of work (and it's God's work) you're putting into this project. My comment-s goal was mostly to encourage the author, because I have been using and testing this feature and it does make a difference. The biggest the VRAM amount with cpu expert offloading scenario, the bigger the reward, for users oj a single gpu (i imagine those are majority) the gains are marginal. Whatever path you take there is a reasoning behind so... I'd say that making this buried somehow would probably make it reach more users anyway and prove its usefulness, and if that's the case perhaps a more streamlined/automated heatmap mapping in the future would make sense as a second step?

DenysAshikhin:
Maybe expert heatmaps are computed automatically during quantization? 
This would be insane if it holds up. have some kind file shipped with the model that shows expert preference.

Also instead of having to compute and store can it instead be dynamic that grows as it is used? Sure it means each restart is bad but as a v1 have it just in memory would be a nicer UX? Then v2 could save a file where model is loaded from of previous values to make it hot right away?

@joaosump 
I think it would be nice if you can explore the things i mentioned and report back here to help narrow down potential options for turbo 🤔

joaosump:
@DenysAshikhin I dug into the idea already actually. One thought I had was that the quantizer could collect router selections while processing its calibration corpus and emit a small profile alongside the quant (not sure how feasible would that be). Profile loads in 0.05 ms so startup overhead would be negligible. One thing that we must take note is that this "heatmap" depends heavily on several factors:

1. Type of decoding: a prose profile applied to code produced only about a 1.01× improvement
2. A profile built from 1K-token windows lost most of its value on long contexts
3. Hardware and amount of vram obviously
4. Other stuff I probably did not measure yet...

For myself using Qwen 3.8 Flash, my "agentic" profile with 238 of 512 resident experts captured about 77.1% of held-out routing, compared with 46.5% for uniform placement. That gave me around 16% uplift in median decoding speed.

You mentioned about this "dyynamical" growth and I think this PR sypports it (seed mode). The current CPU-offload path collects routing activity in memory and dynamically swaps experts. That swapping is slow though, at most 64 experts across all layers are moved every 128 decode steps. On a 42-layer model, that works out to only about 1.5 experts per layer per sweep. Each promotion also rereads the expert weights from the checkpoint, so a session may end before placement has settled.

Persisting the learned counts or ranking across restarts seems like a good next step. I would save the small routing census and reload it as the seed for the next session.

For continuation, I'd go with:

1. Expose dynamic / seed placement as a simple optional default.
2. Accelerate swaps immediately after a cold load.
3. Periodically persist the learned ranking and use it as the next startup seed.
4. Optionally ship a quantization-generated seed when its held-out score demonstrates that the calibration corpus generalizes.

This would give most users a zero-configuration path if they enable that flag while also preserving the larger gains available from purpose-built profiles...

DenysAshikhin:
I know it's not great, but it would help with reproducibility if you could make the save to disk and reload on startup work with the perf.py?

Like say you run it 3,4,5 times. In theory without your PR it would be roughly same speed, but with it, you would see gains in decode per run each time. And if its reproducible by wiping that saved file... that could be a good direction for UX? Dunno, might need some numbers and testing on your side to see how valid it is

midagedev:
FYI: the `can_defer_load` part of this PR (the guard reading `EXL3_MOE_CPU_SPLIT` instead of `infer_params.moe_cpu_split`, so `-mcs` + `EXL3_MOE_CPU_SPLIT_STATS` generated garbage) bites on the current release on its own, so I split just that fix out as #376 with a reproducer and a unit test. If this PR lands first, #376 is subsumed by it — no change to what you have here.

eMarco:
(disclaimer: pardon my laziness, this comment is written by GPT Sol and is extracted from a series of experiments I ran to fine-tune Qwen 3.8 FN on a server grade machine. Of course, I have read it)


I have some Qwen3.8-Flash-Next results that may be useful for the reproducibility / startup-placement discussion above.

I retested the PR on post-1.5 `dev` (`903b216`). I used the same build and runtime settings throughout; the comparison is simply no profile vs the same runtime initialized from a precomputed profile in `seed` mode.

Setup:

- Qwen3.8-Flash-Next EXL3, 4.05 bpw
- NVIDIA A40 48 GB, PCIe Gen4 x16
- measured pinned H2D bandwidth: ~25.9 GB/s
- 2× EPYC 7742 host; 96-vCPU XCP-ng guest, GPU passthrough, no guest vNUMA (due to a XCP-ng limitation unfortunately)
- CPU pinning disabled
- `--moe_cpu_split 250` (262/512 routed experts resident per layer)
- 16 CPU-MoE threads (grid selected)
- pinned arena enabled
- default dynamic swapping
- dynamic MTP, max depth 3 (again, grid selected)

The profile was built from separate coding/Agent sessions using roughly 48k-context windows. The test sequence was not part of the profile data: nine frozen VSCode Agent requests, with prompt contexts around 44k-51k tokens.

Three paired runs, restarting the model for every measurement and alternating the order:

| run | no profile | profile seed | delta |
| --- | ---: | ---: | ---: |
| 1 | 28.215 t/s | 51.308 t/s | +81.85% |
| 2 | 28.725 t/s | 49.671 t/s | +72.92% |
| 3 | 28.653 t/s | 50.270 t/s | +75.44% |

Median paired improvement: **+75.44%**.

I also ran one diagnostic comparison with `EXL3_MOE_HANDOFF_PROF=1`:

| | no profile | profile seed |
| --- | ---: | ---: |
| `cpu-assign/row` | 4.083 | 1.319 |
| cold routed experts | 51.03% | 16.48% |
| profiled decode | 28.762 t/s | 50.439 t/s |

So the profile reduced cold expert assignments by about 67.7%. The profiled throughput result is consistent with the unprofiled runs above, but I use the latter for the performance figure.

Median MTP acceptance in the unprofiled runs went from 59.65% to 62.45%. Cold-prefill throughput was about 3.3% lower with the profile; I haven't isolated that difference.

I used `seed` because it seems closest to the persisted-ranking idea discussed above: start from a known placement and then let the existing online swapper continue adapting. I also used `static` while validating the profile, mainly as a way to separate profile quality from online adaptation. I haven't repeated static-vs-seed on the current post-1.5 runtime, so I don't have a current number for that comparison. 

I ran into the reproducibility issue mentioned above as well. I don't think `perf.py` is a good fit for this particular measurement for the reason already described in the PR: synthesizing recurrent state is not equivalent to accumulating a real long context through the Generator, and the routing changes materially with context.

For the measurements above I ended up freezing a real request sequence and replaying it through the normal serving/Generator path. The harness fixes request order and seeds and hashes the captured inputs, then runs paired configurations against fresh model loads.

The capture side still needs some cleanup before I would call it portable. In particular, saving the OpenAI request alone isn't sufficient when sampler or model defaults may be supplied by the server. I'm going to update it to record the resolved generation configuration as well and to capture/replay through TabbyAPI.

A small real-sequence replay benchmark may be useful here independently of this PR: capture/tokenize a sequence once, then replay the same contexts and sampler settings through `Generator`, reporting throughput and `cpu-assign/row`. That would also give a reproducible way to test the persistence idea above: cold dynamic placement, persist what it learned, reload it on subsequent runs, then delete the saved state and verify that performance returns to the cold baseline.

If that sounds useful, I can clean up the replay part into a separate PR rather than adding more machinery to this one.

joaosump:
> I know it's not great, but it would help with reproducibility if you could make the save to disk and reload on startup work with the perf.py?
> 
> Like say you run it 3,4,5 times. In theory without your PR it would be roughly same speed, but with it, you would see gains in decode per run each time. And if its reproducible by wiping that saved file... that could be a good direction for UX? Dunno, might need some numbers and testing on your side to see how valid it is

Hard to say, I don't think of it as "you would see gains in decode per run each tim" but rather as a use the model for a bit, profile is adjusted and saved in to a static best for your workload. Defo would not make this a default enabled flag and make that optional. Sorry, I'm on vacation but once I'm back in the weekend I will dig in a bit more :)

PR #316, opened 2026-09-01
attention_fn: Route non-power-of-2 head dims away from flash-attn

whoisjeremylam:
AI disclosure: I'm a programmer but not one familiar with ML. GLM 5.3 and DS4 Flash were used in the investigation.

For head dims that are not a power of two, the flash-attn 2 kernels read uninitialized memory. This causes nondeterministic outputs, can contain NaN, which poisons the KV cache.

This affects the Qwen3-VL vision tower (`head_dim = 72`) and shows up as degenerate output (repetition loop of `!`) after image input. I had this behavior when testing Qwen3.8-27B with Pi Coding Agent.

Here is the minimal code to reproduce the issue (flash-attn 2.8.3, sm_86):

```python
import torch
from flash_attn import flash_attn_func

q = torch.randn(1, 512, 16, 72, dtype=torch.half, device="cuda")
k = torch.randn(1, 512, 16, 72, dtype=torch.half, device="cuda")
v = torch.randn(1, 512, 16, 72, dtype=torch.half, device="cuda")
outs = [flash_attn_func(q, k, v, causal=False, softmax_scale=72 ** -0.5) for _ in range(4)]
print(max((outs[0] - o).abs().max().item() for o in outs[1:]))
# ~0.5 (should be 0.0); head_dim 64 gives exactly 0.0
```

The fix routes non-pow2 head dims away from flash-attn and into the existing deterministic fallbacks:

- `flash_attn_2.py`: all three entry points (`fn_flash_attn_func`, `fn_flash_attn_with_kvcache`, `fn_flash_attn_varlen_func`) decline non-pow2 head dims via a new `_is_pow2` check
- `triton_paged.py`: `fn_triton_paged_attn` declines non-pow2 as well — its kernel raises `ValueError` on them (the other Triton paged wrappers already gate on this), and it sits before the fallbacks in dispatch order
- `torch.py`: `fn_torch_sdpa_fallback_cache` accepts non-pow2 below 512 so cached non-pow2 calls still have a backend (pow2 stays on the fast paths)

Based on testing (and also matches the Triton kernels' existing requirement), the required condition is power of 2 rather than `dim % 32` because head_dim 160 crashes flash-attn. with an illegal memory access.

`tests/test_non_pow2_head_dim.py` covers the added condition and verifies the fallbacks produce deterministic, correct output for `head_dim = 72`.

### DIFF
--- a/exllamav3/modules/attention_fn/flash_attn_2.py
+++ b/exllamav3/modules/attention_fn/flash_attn_2.py
@@ -9,13 +9,19 @@ try:
 except (ModuleNotFoundError, ImportError):
     has_fa2 = False
 
+
+def _is_pow2(n: int) -> bool:
+    return (n & (n - 1)) == 0 and n > 0
+
+
 def fn_flash_attn_with_kvcache(args: AttnArgs) -> torch.Tensor | None:
     if (
         not has_fa2 or
         args.sinks is not None or
         args.is_varlen() or
         not args.has_kv_cache() or
-        args.dim > 256
+        args.dim > 256 or
+        not _is_pow2(args.dim)  # FA2 requires power-of-2 head dims
     ):
         return None
 
@@ -46,7 +52,8 @@ def fn_flash_attn_func(args: AttnArgs) -> torch.Tensor | None:
         args.is_varlen() or
         args.has_kv_cache() or
         args.dim > 256 or
-        args.non_causal_spans
+        args.non_causal_spans or
+        not _is_pow2(args.dim)  # FA2 requires power-of-2 head dims
     ):
         return None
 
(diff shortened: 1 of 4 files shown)


turboderp:
I think you're seeing some other bug here, because FA2 should support all head dims divisible by 8. There's no pow2 requirement. I also can't reproduce the error here. 

Is it possible you have a bad FA2 wheel installed? Wrong architecture running through PTX-JIT or something?

It is true, though, that the Triton kernels should decline rather than raise, since otherwise there's no fallback when FA2 isn't installed.

whoisjeremylam:
Thank you for taking the time to reply and correcting my knowledge. I'll look into the build.

whoisjeremylam:
Just following up to mention that you were correct about the bad wheel. In a fresh venv using the `kingbri1/flash-attention` v2.8.3 wheel, I can reproduce the error. Compiling FA2 from source has no such problems.

Issue #318, opened 2026-09-01
Per-job failures latch the AsyncGenerator and strand recurrent-state slots until the server wedges permanently


creslinux:
## Summary

In 1.4.5, an exception raised while a job is being started or prefilled (a CUDA OOM during prefill, an allocation assert, any bug in that path) escapes `Generator.iterate()` unguarded and lands in `AsyncGenerator._run_iteration`, which sets `self.error`, ends the one-and-only iteration task, and errors every in-flight job. The wrapper never recovers: `enqueue()` rejects every future job with the stored exception.

```python
# async_generator.py — _run_iteration
except Exception as e:
    self.error = e
    for async_job in self.jobs.values():
        async_job.put_result(e)
    self.jobs.clear()

# enqueue()
if self.error is not None:
    job.put_result(self.error)   # every future job, forever
    return
```

`self.error` is set once and never cleared; `iteration_task` is created once in `__init__` and never recreated.

## Downstream self-healing hides this, until it doesn't

TabbyAPI catches the error at the consumer and calls `create_generator()`, which replaces the `AsyncGenerator` and the sync `Generator` — but reuses the **same `Cache`**, and the `Cache` owns the recurrent-state slot pool (`num_slots = max_batch_size`; `Cache` defaults to 16, TabbyAPI configures 4 for recurrent models, which is how we hit it). The failed job is discarded without `deallocate_pages()`, so its slot is stranded in `free_list` permanently — along with the slot of every other job that was in flight at that moment, since after the latch `AsyncGenerator.cancel()` deliberately does not touch the sync generator.

Each incident strands at least one slot. At pool size, every subsequent request dies on `Cannot create new state: no available slots`, and no amount of generator recreation can refill a shared, drained cache. Only a process restart recovers. Meanwhile `/v1/models` keeps returning 200 and the weights stay resident, so port- and model-list-based monitoring sees a healthy server.

## Demonstrated by fault injection

One `raise` in `Job.prefill`, gated on prompt length; every handling path exercised is unmodified 1.4.5. Pool = 4 slots. Sequential incidents, one request at a time:

| | stock 1.4.5 | with per-job containment (#319) |
|---|---|---|
| incidents 1–3 | request errors; next request served (recreation heals) | same; error delivered to that request only |
| incident 4 | **permanent wedge** — `no available slots` on every request after; restart required | contained, serving |
| incidents 5–7 | wedged | contained, serving |

Concurrency (one healthy request in flight alongside an incident):

[...]

PR #319, opened 2026-09-01
Contain per-job failures in Generator.iterate

creslinux:
Addresses #318.

`iterate_start_jobs` (via `job.allocate_pages()`) and the `job.prefill()` loop in `Generator.iterate` run unguarded. Any per-job exception there escapes to `AsyncGenerator._run_iteration`, which latches `self.error` permanently and ends the iteration task for every job (mechanism in #318). Observed in production serving as: one CUDA OOM during prefill, then every subsequent request failing until restart, while `/v1/models` keeps returning 200.

## Change

- New `Generator.reap_failed_job(job, error, results)`: releases the failed job's pages and recurrent state, drops it from `active_jobs`, and emits an error result.
- Guards at both sites (`allocate_pages` in `iterate_start_jobs`, and the prefill loop). On failure the job is reaped and the iteration continues with the remaining jobs.
- Error results carry the full result contract (`serial`, `stage: "error"`, `eos: True`, `error`). `enqueue()`'s docstring documents `serial` as present on started/prefill/streaming results, and the sync `generate()` reads all three keys.
- Async side: `deliver_results()` (extracted from `_run_iteration` for testability) queues the raw exception object for that job only, so `AsyncJob.__aiter__` re-raises it. Deliberately not a normal result dict — a dict with `eos: True` would stream to the consumer as a clean, empty completion.
- Sync side: `generate()` raises `r["error"]` on `stage == "error"` instead of treating it as a clean finish and returning a silently truncated completion.
- `deallocate_pages()` failing during reaping is logged with `exc_info` rather than swallowed.

## Deliberately not changed

Generation-stage failures remain unguarded and still reach the latch. Full coverage of `iterate_gen`'s shared-batch internals is a larger change, and post-containment the latch firing is a useful signal that a site was missed. Whether the latch should ever auto-recover is left as the design question in #318.

## Evidence

Fault-injection A/B (one `raise` in `Job.prefill` gated on prompt length; 4-slot pool; tables in #318): stock 1.4.5 wedges permanently at exactly pool-size incidents; this branch contains 7/7 incidents with the server serving after each.

Scope note: this PR alone does not protect *concurrent* requests when serving through TabbyAPI, because its consumer handler recreates the generator on any error and cancels in-flight jobs. That needs a companion downstream change (recreate only when `generator.error` is set), which will be submitted separately. The library-level guarantee — the generator keeps serving other jobs — is covered by `test_failing_prefill_is_contained_in_iterate`.

## Tests

`tests/test_failure_containment.py` — six stub-based tests, no GPU or model required (`Generator` via `__new__` with stub jobs): error-result contract, containment of a failing `allocate_pages` alongside a healthy job, containment of a failing `prefill` inside `iterate`, async delivery as a raw exception, sync `generate()` raising, and cleanup-failure logging.

### DIFF
--- a/exllamav3/generator/async_generator.py
+++ b/exllamav3/generator/async_generator.py
@@ -34,23 +34,7 @@ class AsyncGenerator:
                 # consumer (e.g. a disconnected client that stopped draining its queue) would wedge the whole
                 # generator (issue #227).
                 results = self.generator.iterate()
-                for result in results:
-                    job = result["job"]
-                    async_job = self.jobs.get(job)
-                    if not async_job:
-                        continue
-                    if result.get("stage") == "error":
-                        # Per-job failure contained by Generator.reap_failed_job. The raw
-                        # exception must go into the queue as an exception object, not as a
-                        # result dict: __aiter__ re-raises queued exceptions, which the
-                        # server surfaces as an error for this request. A dict here would
-                        # instead stream as a clean (empty) completion.
-                        async_job.put_result(result["error"])
-                        del self.jobs[job]
-                        continue
-                    async_job.put_result(result)
-                    if result["eos"]:
-                        del self.jobs[job]
+                self.deliver_results(results)
 
                 # Yield back to the event loop so result consumers and cancellation requests can run between
                 # generator iterations, even if the synchronous generator immediately has more work available.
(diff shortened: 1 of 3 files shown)


turboderp:
Looks good :+1:

Issue #162, opened 2026-03-04
Please consider keeping older GPU support (SM75/T4) in the roadmap


mikehalow8080-netizen:
Hi turboderp,

First — thank you. ExLlamaV2 changed what's possible for people like me. I'm an electrician in Indiana. I'm not a tech company. I built a server out of a used Dell R740xd with three Tesla T4s I could afford, and because of your work, I run a fine-tuned 32B model on that hardware every day. It works. It's real. Your engine made that possible.

I'm writing because ExLlamaV3 appears to be moving toward newer GPU architectures, and I understand why — the performance gains on Ampere and Hopper are significant, and that's where the cutting edge is. But I want to ask you to consider the people who can't follow that curve.

Tesla T4s are SM75. They're old. They're also $700+ on eBay now. For a lot of us — hobbyists, independent developers, people in rural areas, people in countries where an RTX 4090 costs six months of wages — older GPUs are the only path to running these models locally. The whole point of local inference is independence. If local inference requires a $1,600 GPU, that independence belongs to the same people who already have access to everything.

I'm not asking you to optimize for old hardware. I'm asking you to not lock us out of it. If SM75 support means a slower codepath, that's fine. We'll take slower. What we can't take is "not supported." The difference between slow and impossible is everything.

The models are getting better and more accessible every month. Qwen 3.5 27B matches Claude-level reasoning and it's Apache 2.0. The model side of this is opening up. If the inference side closes down to only the newest hardware, the people who benefit most from open models lose access to them at the exact moment they become worth running.

Your work has always been about making large models accessible on consumer hardware. That mission matters more now than it did when you started. The models have caught up to the commercial APIs. The only question left is who gets to run them.

Thank you for everything you've built. I hope you'll keep people like me in the room.

Mike
Clinton, Indiana

Issue #81, opened 2025-09-06
T4 Nvidia Google Colab support


ClevyPasserby:
Hello, I'd like to request support for Colab T4 Nvidia. Exllamav2 works fine but not Exl3. I don't have powerful computing resources for LLMs. Will wait for the support.

turboderp:
Yeah this can't really work on Colab T4. I'm planning to add some stuff to make it easier to get up and running on RunPod, or maybe other similar services.

PR #325, opened 2026-09-02
Turing (sm_75) support

rluisr:
Adds Turing (sm_75) support, per #162 and #81. On a T4 today the build dies at PTX assembly:

```
ptxas error : Feature 'cp.async' requires .target sm_80 or higher
ptxas error : Feature '.m16n8k16' requires .target sm_80 or higher
```

The ask in #162 was explicitly "we'll take slower, what we can't take is not supported", and that is what this is. It is not competitive with Ampere and does not try to be.

**Result: DeepSeek-V4-Flash-0731-exl3 2.77bpw runs on 4x Tesla T4.**

Based on `dev` @ 843725c. ~185 lines of functional change; the rest is docs and three tests.

## Re: the three blockers in #162

Your assessment was that the GEMM kernel would need a complete redesign. Two of the three points turned out narrower than that, and I'd appreciate scrutiny on whether I'm wrong about them.

### 1. `mma.m16n8k16` → two `mma.m16n8k8`

You wrote that "the lane mappings for m16n8k8 are different so you can't just run two of those mmas instead."

For the `.row.col` layout EXL3 uses, I believe they do line up. In m16n8k16 the A fragment is `{a0,a1,a2,a3}` where `{a0,a1}` covers k=0..7 and `{a2,a3}` covers k=8..15, with the same lane-to-element mapping m16n8k8 uses for its two registers; B splits into `b0`/`b1` and C is identical in both. Chaining two k=8 mmas through one accumulator computes the same products over the same operands.

Not bit-identical, and I'd overstated this in an earlier revision of the branch: PTX doesn't specify the internal accumulation order of a k=16 step, and splitting forces one extra rounding of the partial sum at the k=8 boundary. That's far below the quantization error already present at 2-4 bpw, and the tests bound it against a dequantize-then-matmul reference rather than asserting equality.

Sites: `ptx.cuh` (both accumulate variants) and `exl3_gemv_kernel.cuh` (`mma_ab_h` carries its own inline asm).

### 2. `cp.async` → synchronous 16 B copies

`cp_async()` degrades to a load/store pair through registers; `cp_async_fence()`/`cp_async_wait()` become no-ops.

This holds because the pipeline never relied on `cp.async.wait_group` alone for ordering: `wait_stage()` in `exl3_gemm_inner.cuh` issues a `__syncthreads()` after every wait, and a synchronous store has retired by then. The cost is the lost global→shared overlap. `test_gemm_deterministic` guards the assumption — a race would surface as run-to-run variation rather than a wrong-but-stable answer.

### 3. Shared memory

`SMEM_MAX` still bounds the compile-time `static_assert`, but what gets *requested* is now per-device via `DevCtx::get_smem_request()`. Shapes that don't fit are filtered out of kernel selection instead of failing at launch. On Turing only GEMM shape 4 at 8 bpw (66 KB) is excluded; every MoE instantiation fits in 44 KB.

The footprint is derived rather than tabulated: `exl3_gemm_smem_bytes()` takes the `EXL3_GEMM_SHAPE_n` expansion (where `SH_STAGES` is already declared) and reproduces the layout `exl3_gemm_kernel_inner` builds — and that kernel `static_assert`s per instantiation that the two agree, so drift between the filter and reality is a compile error, not an out-of-bounds read past the `extern __shared__` block. (An earlier revision hand-maintained a `{0,6,4,4,4}` table; verified the derived values are identical across all 32 shape/bitrate combinations.)

## Two more that only showed up on hardware

**Paths that abort rather than degrade.** The int8 GEMV sizes its shared memory against a fixed ~80 KB budget that the *device* code also depends on (the kernel recomputes `rows_per` from the same constants), so the host can't request less without desyncing the two — and `cudaFuncSetAttribute` failing there hits a `cuda_check` that kills the process. The quant-direct KV cache path writes K/V into the packed cache *before* dispatching over a Triton-only candidate list, so a backend miss is fatal rather than recoverable. Both are declined on small-smem devices. Quantized caches still work; they just materialize fp16 temporaries.

**Three attention paths with baked-in tiles.** `triton_paged.py` and `dsa_triton.py` now size tiles per device — at DeepSeek's `D_c = 512` the KV tiles dominate, so `block_n` has to come down too, or the one-shot DSA kernel asks for 140 KB. The batch-compiled AOT kernels (`bc_attn.py`/`bc_dsa.py`/`bc_mla.py`) have `constexpr` tile shapes that can't be resized, and their failure mode is an opaque `RuntimeError: CUDA driver error` out of `CU_FUNC_ATTRIBUTE_MAX_DYNAMIC_SHARED_SIZE_BYTES`; `_compile_kernel` now checks the compiled requirement and raises `BCKernelTooLarge`, which the builders catch to decline to the eager path.

Since Triton raises `OutOfResources` rather than recompiling smaller, the dispatcher entry points also treat that one error as a dispatch miss and fall through to SDPA/xformers.

## Verified on hardware

Tesla T4 (sm_75, `smem_optin=65536`), CUDA 13.0, torch 2.9.1+cu130, this branch:

| | Result |
|---|---|
| `tests/test_sm75_gemm.py` | 15 passed |
| Llama-3.2-1B-Instruct-exl3 4.0bpw (dense) | 134 tok/s, coherent |
| Qwen3-30B-A3B-exl3 3.0bpw (MoE) | 42.5 tok/s, coherent |
| DeepSeek-V4-Flash-0731-exl3 2.77bpw | 5.3 tok/s, coherent |

DeepSeek-V4 is 93 GiB of weights on 4x15 GiB with 40 MoE layers offloaded to system RAM, exercising MLA, DSA sparse attention and the CPU expert handoff:

```
The three primary colors are red, blue, and yellow. In additive color
models (like RGB), the primary colors are red, green, and blue.
```

`m` is swept 1/8/16/32 in the unit tests deliberately: m=1 takes the GEMV path, m≤8 the small-m reduction, m>16 the multi-pass loop, and each drives the mma differently, so a wrong fragment mapping wouldn't necessarily fail all of them.

### No Ampere regression

Also built and run on an RTX 3090 (sm_86, CUDA 12.4, torch 2.6), at the same final commit:

| Check | Result |
|---|---|
| `tests/test_sm75_gemm.py` | 15 passed |
| `g_get_cc` | 2 (`CC_AMPERE`, not `CC_OLD`) |
| `g_get_smem_max` | 101376 — raw driver value, unchanged |
| `_dev_small_smem` | False — tiles unshrunk, quant-direct cache and int8 GEMV both still active |

Every PTX substitution is behind `#if EXL3_SM75` (`__CUDA_ARCH__ < 800`), so an sm_80+ cubin is byte-identical to what it is today.

Worth flagging that an earlier revision of this branch *did* regress Ampere: `get_smem_max()` clamped the driver value to `SMEM_MAX`, which made sm_86 report 92160, fall under a 96 KB threshold, and take the Turing paths. Fixed by splitting the two meanings (`get_smem_max` = capability, `get_smem_request` = what a kernel may ask for) and keying the Python gates on compute capability `< 8` instead — Volta reports exactly 96 KB, so no byte threshold separates it from Ampere's 99 KB robustly.

## Cost on Turing

- No `cp.async`, so global→shared traffic no longer overlaps with compute
- Two mma instructions per tile instead of one
- 64 KB of shared memory caps tile sizes and occupancy
- No flash-attn; attention runs Triton or SDPA
- int8 GEMV and the batch-compiled (CUDA-graph) attention paths are declined

## Building

`TORCH_CUDA_ARCH_LIST` must include 7.5 — the release wheels contain no sm_75 cubin, and I have not touched the wheel matrix in this PR.

```bash
TORCH_CUDA_ARCH_LIST=7.5 MAX_JOBS=24 pip install --no-build-isolation -e .
```

## Points I'm least sure of

1. The m16n8k16 → 2x m16n8k8 fragment mapping, especially `mma_ab_h` which passes A as two `FragB` halves. If the lane mapping argument is wrong, this is the thing that's wrong.
2. Whether the synchronous `cp.async` fallback is genuinely race-free in every path, including `exl3_gemv_int8_kernel.cuh`'s staged unit. The claim rests entirely on existing `__syncthreads()` placement.
3. `exl3_gemm_shape_compat()` is now device-dependent but takes no device argument — it reads the current CUDA device. Every in-tree caller runs under a device guard, but it's a sharp edge on a mixed-arch host, and I'd rather thread a device through if you prefer.
4. Whether declining the int8 GEMV wholesale on small-smem devices is the right call versus making `gemv_int8_sq_rows_max()` device-aware. The latter is better but needs the kernel and host to agree on a runtime value, which is a larger change than I wanted to make unreviewed.

Happy to split this into separate PRs (PTX fallbacks / shared-memory sizing / attention tile sizing) if that's easier to review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
 
-# <img src="doc/cat.png" width="40"> ExLlamaV3
+# <img src="doc/cat.png" width="40"> ExLlamaV3 (Turing / sm_75 fork)
+
+> **This fork adds Turing (sm_75) support.** Upstream EXL3 requires Ampere or newer; see
+> [upstream #162](https://github.com/turboderp-org/exllamav3/issues/162) and
+> [#81](https://github.com/turboderp-org/exllamav3/issues/81). Tesla T4, RTX 20-series and
+> Quadro RTX now run EXL3. See [doc/sm75.md](doc/sm75.md) for what was changed and what it
+> costs. Everything below is upstream documentation.
 
 ExLlamaV3 is an inference library for running local LLMs on modern consumer GPUs. Headline features:
 
--- a/doc/sm75.md
+++ b/doc/sm75.md
@@ -24,9 +24,12 @@ same lane-to-element mapping m16n8k8 uses for its two registers; B splits the sa
 `b0` and `b1`, and the C fragment is identical in both. Chaining two k=8 mmas through one
 accumulator therefore computes the same sum over the same k range.
 
-This is bit-equivalent, not an approximation — accumulation order within a k=16 step is not
-observable, because each mma internally accumulates in a fixed order anyway. The unit tests
-assert against a dequantize-then-matmul reference to confirm this empirically.
+This is not guaranteed bit-identical. PTX does not specify the internal accumulation order of
+a k=16 step, and splitting forces the partial sum to be rounded at the k=8 boundary where the
+fused instruction need not round. The bound is one extra rounding per 16-element dot product,
+orders of magnitude below the quantization error EXL3 already carries at 2-4 bpw. The unit
+tests bound the end-to-end deviation against a dequantize-then-matmul reference rather than
+asserting equality.
 
 Sites: `exllamav3_ext/ptx.cuh` (both fp32- and fp16-accumulate forms) and
 `exllamav3_ext/quant/exl3_gemv_kernel.cuh` (`mma_ab_h`, which has its own inline asm).
@@ -59,6 +62,22 @@ it raises `OutOfResources` at launch instead of recompiling smaller. Tile config
 chosen against the real device limit, and the dispatcher entry points treat a
 shared-memory `OutOfResources` as a dispatch miss, so an unanticipated geometry falls through
 to SDPA/xformers instead of killing the forward pass.
+
+### 4. Paths disabled on small-smem devices
+
+Two optimizations size their shared memory against fixed ~80 KB budgets that the *device* code
+also depends on (the kernels recompute their own row counts from the same constants), so the
+host cannot request less without desyncing the two. Both are declined on Turing:
+
+- **int8 GEMV** (`EXL3_INT8_GEMV`, default on): a decode-time alternative to the fp16
+  tensor-core kernel, which handles the same work. Declining costs throughput only.
+  Without this gate `cudaFuncSetAttribute` fails and the following `cuda_check` aborts the
+  process.
+- **Quant-direct KV cache attention**: this path writes K/V into the packed cache *before*
+  dispatching, and then dispatches over a Triton-only candidate list, so a backend miss is
+  fatal rather than recoverable. On Turing the dispatcher takes the dequantize-then-attend
+  path instead, which can fall through to SDPA/xformers. Quantized caches still work; they
+  just materialize fp16 temporaries.
 
 ### Not changed
 
--- a/doc/sm75.md
+++ b/doc/sm75.md
@@ -52,10 +52,17 @@ unconditionally, which fails the launch outright.
 
 `SMEM_MAX` still bounds the compile-time `static_assert` (every template instantiation must
 keep compiling), but what gets *requested* is now per-device, from
-`DevCtx::get_smem_max()` &rarr; `cudaDevAttrMaxSharedMemoryPerBlockOptin`. Shapes that do not
-fit are filtered out of kernel selection by `exl3_gemm_shape_compat()` rather than attempted
-and failed. In practice only GEMM shape 4 at 8 bpw (66 KB) is excluded on Turing; every MoE
-instantiation fits in 44 KB and is unaffected.
+`DevCtx::get_smem_request()` &rarr; `cudaDevAttrMaxSharedMemoryPerBlockOptin` capped at
+`SMEM_MAX`. Shapes that do not fit are filtered out of kernel selection by
+`exl3_gemm_shape_compat()` rather than attempted and failed. In practice only GEMM shape 4 at
+8 bpw (66 KB) is excluded on Turing; every MoE instantiation fits in 44 KB and is unaffected.
+
+The footprint itself is derived, not tabulated: `exl3_gemm_smem_bytes()` in
+`exl3_kernel_map.cuh` takes the `EXL3_GEMM_SHAPE_n` macro expansion (where `SH_STAGES` is
+already declared) and reproduces the layout `exl3_gemm_kernel_inner` builds. That kernel
+`static_assert`s, per instantiation, that the shared function agrees with its own layout, so
+a divergence between the filter and the reality is a compile error rather than an
+out-of-bounds read past the `extern __shared__` block.
 
 The Triton attention kernels have the same problem and Triton does not degrade gracefully —
 it raises `OutOfResources` at launch instead of recompiling smaller. Tile configs are now
--- a/doc/sm75.md
+++ b/doc/sm75.md
@@ -79,6 +79,24 @@ host cannot request less without desyncing the two. Both are declined on Turing:
   path instead, which can fall through to SDPA/xformers. Quantized caches still work; they
   just materialize fp16 temporaries.
 
+### 5. Attention kernels sized per device
+
+Three separate attention paths bake in tile shapes chosen for ~100 KB of shared memory:
+
+- **`triton_paged.py`** (paged prefill/decode): tile configs now shrink on small-smem devices,
+  and the dispatcher entry points treat a shared-memory `OutOfResources` as a dispatch miss,
+  falling through to SDPA/xformers.
+- **`dsa_triton.py`** (DeepSeek sparse attention): at `D_c = 512` the KV tiles dominate
+  (`block_n x 512 x 2 B` per stage), so `block_n`, the head tile and the pipeline depth all
+  come down. Without this the one-shot kernel asks for 140 KB.
+- **`bc_attn.py` / `bc_dsa.py` / `bc_mla.py`** (batch-compiled AOT kernels): tile shapes are
+  fixed `constexpr`s, so these cannot be resized. The cubin still compiles on Turing, but
+  `CU_FUNC_ATTRIBUTE_MAX_DYNAMIC_SHARED_SIZE_BYTES` then fails with an opaque
+  `RuntimeError: CUDA driver error`. `_compile_kernel` now checks the compiled kernel's
+  shared-memory requirement against the device and raises `BCKernelTooLarge`, which the BC
+  builders catch to decline to the eager path — the same path that does size its tiles per
+  device.
+
 ### Not changed
 
 `ldmatrix`, `__nanosleep`, `__dp4a`, `tanh.approx.f32` and `cuda::atomic_ref` are all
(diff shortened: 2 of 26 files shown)


weicj:
Validated on **dual 2080Ti 22G** with [2.05 bits per weight](https://huggingface.co/turboderp/Qwen3.8-Flash-Next-exl3/tree/2.05bpw_h4_ng4) of [Qwen3.8-Flash-Next-exl3](https://huggingface.co/turboderp/Qwen3.8-Flash-Next-exl3).  tensor_p is not supported by current implementation of qwen4exp arch with PLE SSD off-load/streaming, so I tested with autosplit (layer-split).

Under 4096 PP / 128 TG, I got: **prefill 547.09 tok/s** and **decode 39.71 tok/s**.  

For layer-split mode, decoding speed is impressive (way better than I thought), prefill is usable but slower than I expected. So I've tried a quick integration of **FlashInfer** to make FlashAttention-2 support. No luck, for the same pp/tg, I've only had 5% improvement to hit **577.08 tok/s** prefill with fa2. There seems to be a bigger bottle neck than dense prefill.

vcruz305:
I have been doing the same port independently on a Quadro RTX 6000, so here is a third Turing part as a data point, plus notes on two of your open questions.

Caveat first: these numbers come from my own branch, not this one. I landed on the same three substitutions but handled shared memory differently, so treat this as corroboration rather than validation of your PR. I am happy to build this branch on the Quadro and report back properly if that is useful.

**Quadro RTX 6000, 24 GB, sm_75, 64 KB smem. Qwen3.8-27B EXL3 4.00bpw, Q4 KV cache, single card, TabbyAPI.**

| | |
|---|---|
| decode | 31 to 33 tok/s |
| decode, MTP at ndt=1 | 42 to 45 tok/s |
| prefill | 812 tok/s at 24k, 457 at 32k, 296 at 64k |
| context | 262,144 allocated, 19.7 GiB |
| two concurrent streams | 49.9 tok/s aggregate |

Long context is the thing I have not seen covered in this thread. The full 262k window allocates and serves. I have checked retrieval to 64k depth and no further, so that is as far as I would claim. It also scores where I would expect it to on a small reasoning eval I use, which is worth more to me than a coherence check, since a subtly wrong fragment mapping tends to survive "does it produce English" and not survive scoring.

**On your point 1, the m16n8k16 to two m16n8k8 mapping.** I reached the same conclusion for the `.row.col` case independently, including that `mma_ab_h` in `exl3_gemv_kernel.cuh` carries its own inline asm and is a third site separate from the two in `ptx.cuh`. Two people converging is not proof, but I got there without seeing your branch, and I sweep m the same way you do for the same reason.

**On your point 2, `exl3_gemv_int8_kernel.cuh`.** I patched that file rather than declining int8 GEMV, and it has not produced garbage for me. I would not offer that as evidence of race freedom though. A race that resolves correctly in the common case is precisely what survives that kind of testing. Declining it on small-smem devices is the more conservative call and I think it is the right one for a first merge.

One place your approach is plainly better than mine: I gated on a compile-time `EXL3_SM75` macro derived from `TORCH_CUDA_ARCH_LIST`, which cannot express a mixed-arch host and quietly produces the wrong kernels if the arch list is wrong. Runtime per-device sizing through `DevCtx` is the right shape, and the `get_smem_max` versus `get_smem_request` split is a fix for a class of bug my version cannot even represent. Reading your note about the earlier revision clamping the driver value and pushing sm_86 onto the Turing paths was useful, since that trap is invisible from where I was standing.

My patches are at https://github.com/vcruz305/exl3-turing for comparison. I do not think there is anything in there this PR does not already cover more thoroughly, and I would rather see this merge than split the effort.

rluisr:
Thanks — a third Turing part and an independent arrival at the same three substitutions is worth more than another T4 run would be, and the long-context numbers cover the gap I couldn't: 262k allocated and served, retrieval to 64k. My DeepSeek-V4 run only demonstrates that MLA and DSA don't produce garbage, not that they hold up at depth.

I went through your patches against this branch. Pushed one change (9f811ba); the rest I've left, with reasons below in case you disagree.

**What was actually missing: `paged_attn_triton_decode`.**

Your `_clamp_block_n_for_smem` on the decode path caught a real bug in this branch, and it's a nastier one than the prefill case you cite it alongside. The default holds `block_n * head_dim` at 8192, so the KV tiles are 32 KB per stage and 64 KB at the default depth of 2 — a 64 KB device cannot hold them at any `head_dim`. I had shrunk the prefill config and the DSA/MLA tiles but not this one, and the branch's own `_smem_fallback` decorator then hid it: the `OutOfResources` became a dispatch miss and every decode step silently ran on SDPA. That's a fallback doing exactly what it was written to do while masking a case that didn't need it. Now clamped against the device budget before launch, same shape as your version.

Worth noting for anyone reading the thread later: this is why "it produces coherent text" isn't a sufficient bar. Falling back to SDPA is correct, just slower, so nothing about the output would have revealed it.

**What I've left alone, and why.**

`exl3_gemv_int8_kernel.cuh` — you patched it, this branch declines it, and you're right that neither of us has evidence of race freedom. But I don't think the choice hinges on that. The kernel recomputes `rows_per` from the same constants the host sizes shared memory against, so the host can't request less without the two desyncing, and your `EXL3_ROWS_CLAMP` loop shrinks `rows_max` on the host only. `decomp()` does clamp and recompute `ksplit`, so I expect it holds — but it's a host-side value the device derives independently, and I'd want to see that argument made explicitly before shipping it. Declining costs a decode-time optimization over a path that computes the same thing.

`SMEM_MAX` and `EXL3_GEMM_SHAPE_4` — same problem, different mechanism, and this is the one place I'd push back on your approach rather than just noting a difference. Lowering `SMEM_MAX` to 60 KB makes the compile-time `static_assert` reject shapes that would fit on a bigger card, so the *cap* moves when only the *request* needed to. This branch keeps `SMEM_MAX` at the sm_86 value for the assert, derives the actual footprint from the shape macros, and filters non-fitting shapes out of kernel selection at runtime. Same outcome on a T4 — shape 4 at 8 bpw is excluded either way — but it doesn't bake a device's limit into the build.

The `cuda::atomic_ref` → `atomicAdd`/`atomicExch` rewrite I didn't need; C++17 is only a constraint under your toolchain, not upstream's. Your note about keeping `atomicExch` rather than a plain store is correct regardless, and applies to the unmodified upstream code too.

**On the compile-time gate.** Since you flagged this yourself: I'd add that `_archs <= {"7.5"}` also silently disables the accommodations when someone builds `7.5;8.6`, which is the case where a mixed-arch host most needs them. The failure is a launch abort on the T4 in a build that looked fine. Not a reason to redo working patches, but if anyone lifts that `setup.py` hunk it's worth knowing.

If you do build this branch on the Quadro, the number I'd most want is decode tok/s before and after 9f811ba on a model that takes the paged decode path — that isolates the fix rather than the port. A needle run at your 64k depth against the DSA/MLA clamps would also be more than I can currently claim.

DailyDisco:
Would it be possible to run hybrid ampere and turing with 1 model loaded? So that for example 4 3090s and 2 rtx 8000s can be used for exl3?

Ph0rk0z:
I'm in similar boat. 4x3090 and then 2080ti 22g.

PR #330, opened 2026-09-03
[kernel] Add native BF16 I/O for EXL3 decode GEMM and MGEMM

troycheng:
## Background

This came out of a Qwen3.8-27B TP2 decode path where activations stay in BF16. Calling EXL3 at that point adds dtype conversion and layout copies around the GEMM.

ExLlamaV3 already has a per-tensor-N MGEMM that can batch Q/K/V in a single launch, so I benchmarked it first. Different output widths work as expected. The remaining time is spent converting BF16 input to FP16, repeating the input Hadamard transform for each logical matrix, writing separate FP16 outputs, and packing those outputs into the final layout.

The patch removes that surrounding work while leaving the checkpoint format and the existing FP16 Hadamard/GEMM calculation alone.

## Current patch

The branch currently adds three extension entry points:

- `exl3_gemm_bf16_io` accepts BF16 input and writes BF16 output for K5/K6 MCG decode GEMM.
- `exl3_mgemm_bf16_io` batches fixed-width projection slices and can write each result into its final position in one contiguous BF16 output.
- `exl3_mgemm_bf16_io_grouped_had` also shares the transformed input when several matrices use the same input Hadamard vector.

`direct_output` and the terminal synchronization points can be controlled separately. The wrapper checks residency before it allows the terminal barrier to be skipped. Grouped-Hadamard currently requires the whole group to be resident.

With `force_num_sms=0`, the wrapper picks a cooperative grid from the device properties and kernel occupancy. A manual SM count is checked against the device and cooperative residency limits. Shape, buffer, encoded-width, and group-ID errors are reported before the kernel launch. `had_group_ids` is supplied as CPU int32 metadata, checked on the host, and copied by value into the kernel arguments. CUDA Graph capture and changing-input replay both pass with this setup.

I kept the downstream vLLM caller out of this branch. The diff stops at the ExLlamaV3 extension and its tests.

## Operator benchmark

The table covers one projection group from BF16 input to the final contiguous BF16 output. These are historical prototype measurements on `origin/dev` at `f3f7e420`, before the launch-validation changes in `d0094bc`, using Qwen3.8-27B TP2 K5/K6 checkpoint tensors from both ranks. The machine used an RTX 5090 (SM120), PyTorch 2.13.0+cu130, and CUDA 13.2. Inputs change between CUDA Graph replays, and the benchmark rotates about 256 MiB of weights to keep the working set out of L2. Each result comes from nine alternating samples with 64 graph replays per sample.

| projection | M | fixed-width grouped | current per-N | time saved |
|---|---:|---:|---:|---:|
| GDN Q/K/V/Z | 4 | 26.87 us | 45.24 to 45.27 us | 18.36 to 18.39 us |
| GDN Q/K/V/Z | 16 | 28.65 to 28.68 us | 47.21 to 47.25 us | 18.56 to 18.57 us |
| full-attention Q/K/V | 4 | 24.16 to 24.18 us | 51.22 to 51.32 us | 27.06 to 27.14 us |
| full-attention Q/K/V | 16 | 25.90 to 25.92 us | 52.55 to 52.69 us | 26.64 to 26.77 us |

Across the eight cases, with nine paired samples each, the fixed-width path was faster in all 72 pairs. Relative RMS between the two outputs was 0.001276 to 0.001326. The error comes from the FP16/BF16 operation order at the boundary; the projection formula and checkpoint data stay the same.

## Downstream check

I tested an earlier v1.4.6 BF16 prototype end to end in A-B-B-A order with the same vLLM plugin. The wheel SHA-256 is `9ef7719162fecc001cf565cdccc42d0b407c0bf51669be73620c0ae9c2ea0d66`; it predates the launch-validation changes in `d0094bc`. A used the existing FP16 Trellis boundary. B enabled the BF16 input/output paths through the downstream selector. Each run used two RTX 5090 GPUs with TP2 over PCIe/PIX, no NVLink, 1024 input tokens, 256 output tokens, non-speculative decoding, BF16 KV cache, and FULL_DECODE_ONLY CUDA Graphs. These results describe that prototype, not a fresh benchmark of the current PR head.

| concurrency | A mean | B mean | output throughput | mean TPOT | p99 TPOT |
|---|---:|---:|---:|---:|---:|
| c4 | 213.642 tok/s | 238.374 tok/s | +11.576% | -12.737% | -11.232% |
| c16 | 483.874 tok/s | 510.954 tok/s | +5.597% | -6.726% | -5.554% |

The two B runs differed by 0.41% at c4 and 0.10% at c16. A1 and A2 differed by 0.25% and 0.11%. All 8 runs completed. The vLLM selector remains downstream and is not part of this patch.

## Validation

The self-contained test is:

```bash
python3 -m pytest -q tests/test_exl3_bf16_io.py
```

The expanded suite passed all 15 tests on an RTX 5090 with PyTorch 2.13.0+cu130, using the ExLlamaV3 1.4.8 build that carries the two commits from this PR. Its seven BF16 CUDA sources and shared headers match `d0094bc` byte for byte. This follow-up changes tests and documentation only; no kernel was rebuilt or modified.

The tests now compare three changing-input Graph replays against fresh eager results and the FP16-boundary reference. They also cover three distinct Hadamard groups with repeated and reordered IDs, a K5120/N512 bundle at M1/M16 for K5/K6, and untouched trailing output columns. The original direct/scratch, synchronization, launch-selection and invalid-input checks remain. Graph/eager and grouped/ordinary comparisons are exact; the FP16-boundary reference retains the 0.003 relative-RMS tolerance.

The tested 1.4.8 wheel SHA-256 is `90606feea9a28f5423b712bcec3468ea1dc31c18229744dbfb6019e7819f9586`. Buffer shapes, pointer lifetimes and capture constraints are documented in [the API notes](https://github.com/troycheng/exllamav3/blob/codex/sm120-bf16-io-v146/doc/bf16_io.md).

Earlier build checks compiled the full extension and both CUDA translation units for `sm_80`, `sm_86`, `sm_89`, `sm_90`, and `sm_120` with CUDA 13.2. Runtime evidence remains limited to SM120. CUDA 12.8, Windows, and the other compiled architectures still need runtime coverage.

## Relation to the fused-QKV PoCs

Issue #297, PR #299, and `local-inference-lab/vllm#454` take a format-first approach: Q/K/V are jointly quantized into a fused-uniform checkpoint object and loaded by a matching consumer. This branch starts with the existing split payloads and works on the execution around them.

These measurements use the existing split checkpoint payloads. Joint Q/K/V quantization may still have its own quality or allocation value, but that needs a complete producer-to-consumer experiment and was not tested here.

## API question

I have not settled on the API yet. The WIP exposes three separate BF16 functions, but MGEMM already supports per-tensor N. My preference is to add the BF16 boundary, shared input transform, and direct output behavior to the existing GEMM/MGEMM API so the project does not have to maintain two parallel interfaces.

The functions are separate in the WIP because that made the behavior and benchmarks easier to inspect. I am using the Draft PR to settle the interface before cleaning up the final version. The other open question is whether the extension API should land together with an in-tree caller.

I used AI tools during implementation and review. The commands and results above come from the local test runs for this patch.


troycheng:
This follow-up adds stronger Graph replay checks, multiple Hadamard groups and padded-output tests, plus a short API document covering buffer ownership and capture constraints. All 15 tests passed on RTX 5090 using the 1.4.8 build carrying this PR's two commits. The seven relevant CUDA sources and shared headers match the PR head; this follow-up does not change the kernels.

I also corrected the benchmark provenance in the description. The earlier 1.4.6 performance runs used the prototype before the launch-validation changes in `d0094bc`, so I have labeled them accordingly. There is no new performance claim in this update.

The remaining interface question is whether to fold these BF16 options into the existing GEMM/MGEMM entry points. The downstream caller uses the separate functions today. I have kept M24/M32 kernels and framework-specific routing out of this PR.

turboderp:
This seems to combine a bunch of different changes and I'm a little bit confused. Fixed-width groups in the mgemm kernel seem sensible, especially for fused QKV projections with GQA. I tried adding something similar with slicing, and it does result in lower latency for each individual operation but end-to-end the difference is hard to measure.

Fusing BF16 casting into the input and output stages of the GEMM could also make sense, but there aren't any places where that would actually matter in exllamav3. I'm unclear on exactly how this is being incorporated into vLLM.

troycheng:
The vLLM caller is [vllm-mach](https://github.com/troycheng/vllm-mach), an out-of-tree quantization plugin. It loads tensor-parallel EXL3 weight slices and calls EXL3 GEMM/MGEMM from vLLM's linear layers. vLLM still handles model execution, scheduling and the KV cache. For supported decode projections, the BF16 entry points avoid separate casts around those calls when the model activations are BF16. The calls are in [this dispatch code](https://github.com/troycheng/vllm-mach/blob/7cb712466a7755ec58786c51516d7b9ae689ef77/src/vllm_mach/exl3/dense_adapter.py#L1926-L2011). The gains reported in the PR include grouping and output-layout changes as well as BF16 I/O.

Thanks for taking the time to review this. These changes would have been easier to review separately, and I'll keep the follow-up PRs focused. I've seen your [fixed-width slicing implementation in dev](https://github.com/turboderp-org/exllamav3/commit/0431122526bc71e8509027b62b9b808fc775bd96), and I agree that the BF16 interface doesn't currently have a caller in exllamav3 itself. I'll close this PR here and keep the BF16 support downstream.

I've also tested some smaller optimizations in exllamav3's own runtime. I'll submit them separately, starting with the GDN B/A projection launch, with native correctness tests and benchmarks. These changes won't depend on vLLM.

PR #331, opened 2026-09-03
AVX2 MoE GEMV: bytesum-first maddubs accumulate + software prefetch

treo:
The AVX2 tier emulates the VNNI byte-sum with vpmaddubsw, whose i16 pair sums saturate once an activation is inside the pair. The old masked accumulate avoided saturation with 16 ops per row; instead form the product-byte pair sums (b0+b1, b2+b3, each <= 510) once per k-row with x outside the pair, then fold (b0+b1)*x+(b2+b3)*x with one vpmaddwd per token row against a precomputed splat_dup pattern (x8 in both 16-bit slots). 4 shared + 4 per-row ops, +-127 activations, bit-exact vs the masked accumulate (verified K1-K8 x m1-4 against an exact scalar reference). The token loop is unrolled by hand; a runtime-bounded loop spills the pair sums (~1.4x on Zen 3).

Also prefetch the k-major tile stream (row_stride >= 8 KB outruns the HW prefetcher): distance 4 rows, 2 at bits==6 where 4 collapses cold.

Measured (RTX 3090, Ryzen 5900X, 128GB DDR4): decode 30.6 -> 41 tok/s, prefill 872 -> 957 tok/s on Qwen3.8-Flash-Next 2.05bpw MoE-offloaded.

```
python ../exllamav3/eval/perf.py --model_dir models/Qwen3.8-Flash-Next-exl3-2.05bpw_h4_ng4/ --moe_cpu_split 330 --max_length 32768 --chunk_size 2048 -ngr
```
Before:
image
After:
image

For completeness also with the full context (`--moe_cpu_split 380`). 
image

With just 4K context size, for maximum speed (`--moe_cpu_split 256`):
```
python ../exllamav3/eval/perf.py --model_dir models/Qwen3.8-Flash-Next-exl3-2.05bpw_h4_ng4/ --moe_cpu_split 256 --max_length 4096 -cs 4096 --chunk_size 2048 -ngr
```
Before:
image

After:
image

Improvement on the 4bit Quant isn't as big, but still noticeable. Interestingly I get this speed even if I offload fewer layers to the CPU.
```
python ../exllamav3/eval/perf.py --model_dir models/Qwen3.8-Flash-Next-exl3-4.05bpw_h6_ng6/ -mcl 43 --max_length 32768 -cs 32768 --chunk_size 2048 -ngr
```

Before:
image

After:
image

---
Sidenote. This PR was made with the help of Qwen3.8-Flash-Next-exl3-5.05bpw_h6_ng6 running on ExllamaV3 1.4.5 on an RTX Pro 6000. It runs really well on it.

### DIFF
--- a/exllamav3/exllamav3_ext/cpu/moe_mul1.cpp
+++ b/exllamav3/exllamav3_ext/cpu/moe_mul1.cpp
@@ -41,6 +41,12 @@
 // below 2^15, which makes the operand order load-bearing). Accuracy matches the GPU int8-GEMV
 // mode-2 class (~0.9% per-call output RMS). i32 accumulators are safe for k up to ~8192.
 //
+// On the AVX2 tier (no VNNI) the byte-sum must be emulated with vpmaddubsw, whose i16 pair sums
+// saturate once an activation is inside the pair. The accumulate therefore keeps the pair x-free:
+// sum the product bytes once per k-row into i16 pair-sums <= 510, then multiply by x with one
+// vpmaddwd per token row. +-127 activations, bit-exact vs the masked accumulate it replaces,
+// ~2x its throughput (measured on Zen 3 + Zen 5).
+//
 // State extraction uses compile-time (bits, row) index tables for vpermt2var plus immediate
 // funnel shifts, following benchmarks/exl3_cpu_gemm. The GEMV streams k-major with a contiguous
 // band of output tiles held in register accumulators per worker, so cold expert weights are read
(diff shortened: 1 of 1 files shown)


turboderp:
This is great. I also adjusted prefetch on the AVX512 path and gained a boost there. Plus it lead to some profiling revealing a bottleneck in VBMI prefill (unswizzling), so that's addressed too now. :+1:

PR #334, opened 2026-09-04
Add DFlash2 speculative draft support

RodriMora:
Summary

 This adds greedy DFlash2 speculative drafting to ExLlamaV3.

 The implementation targets paged-cache generation with a non-TP target model. It has been tested with:

 - Target: turboderp/GLM-5.3-Flash-exl3-4.05bpw
 - Draft: incoai/GLM-5.3-Flash-DFlash2

 No native extension changes are required.

 What changed

 - Registered the DFlash2DraftModel architecture.
 - Added grouped dynamic causal convolutions around each attention and MLP sublayer.
 - Added BF16 residual normalization. DFlash2 residual values can exceed the FP16 range.
 - Added the DFlash2 candidate selector and greedy chained top-k walk.
 - Preserved target hidden-state taps in checkpoint order.
 - Added bilateral synthetic-block attention for non-causal DFlash2 checkpoints.
 - Corrected the sliding-window conversion from token count to the backend's inclusive left distance.
 - Added cacheless non-causal span support to the Torch SDPA fallback.
 - Added support for optional input_embedding_scale, output_multiplier, and final logit softcapping.
 - Added validation for target compatibility, projection shapes, convolution geometry, draft length, and BF16 hardware support.
 - Required the checkpoint's native seven-token draft window to prevent fixed eight-row draft blocks from exceeding allocated cache pages.
 - Disabled unsupported dynamic draft truncation.

 Verification behavior

 DFlash2 proposals are greedy. The existing target sampler still chooses the final token and accepts draft tokens only while they match the target output. This keeps generation lossless without
 adding a separate stochastic rejection-sampling path.

 Current limitations

 - Tensor-parallel targets are not supported because the selector needs top-k target logits.
 - CFG and other multi-sequence jobs are not supported.
 - The draft length must be exactly seven tokens for the current eight-row checkpoint format.
 - Dynamic draft truncation is not supported.
 - A GPU with native BF16 support is required.

 Tests

 Added 18 focused tests covering:

 - Architecture registration
 - Bilateral and causal mask selection
 - Sliding-window boundaries
 - Cached Triton attention with noncontiguous pages and a page-boundary write
 - Target tap ordering
 - Fixed draft-length validation
 - Dynamic convolution semantics
 - Exact projection widths
 - Candidate chaining
 - Candidate logit scaling and softcapping
 - BF16 residual normalization

 Validation performed:

 ```text
   python tests/test_dflash2.py
   DFlash2 tests passed
 ```

 ```text
   python -m compileall exllamav3 tests/test_dflash2.py
   git diff --check
 ```

 The real checkpoint was loaded in TabbyAPI with a paged FP16 draft cache. Generation completed successfully across repeated speculative rounds, including rejected drafts and cache rollback.

 Follow-up work

 - Tensor-parallel target support
 - Stochastic DFlash2 proposals with rejection sampling
 - Automated full-checkpoint parity tests
 - Compiled-checkpoint round-trip coverage

Note: This work was done with the help of an LLM, but I manually tested and verified it working on my current system, that has a mix of gpus (rtx pro 6000, 5090 and 3090) and dflash2 as the drafter was working properly in my system. 
edit: wrong grammar

### DIFF
--- a/exllamav3/architecture/architectures.py
+++ b/exllamav3/architecture/architectures.py
@@ -7,6 +7,7 @@ from .decilm import DeciLMModel
 from .deepseek_v3 import DeepseekV3Model
 from .deepseek_v4 import DeepseekV4Model
 from .dflash import DFlashModel
+from .dflash2 import DFlash2Model
 from .dflash_laguna import DFlashLagunaModel
 from .dots1 import Dots1Model
 from .ernie4_5 import Ernie4_5Model
@@ -74,6 +75,7 @@ ARCHITECTURES = {
         DeepseekV3Model,
         DeepseekV4Model,
         DFlashModel,
+        DFlash2Model,
         DFlashLagunaModel,
         Dots1Model,
         Ernie4_5Model,
--- a/exllamav3/architecture/dflash.py
+++ b/exllamav3/architecture/dflash.py
@@ -63,6 +63,8 @@ class DFlashConfig(Config):
         self.mask_token_id = self.read_cfg(int, ["dflash_config->mask_token_id", "mask_token_id"], no_default)
         self.target_layer_ids = self.read_cfg(list, ["dflash_config->target_layer_ids", "target_layer_ids"], no_default)
         self.target_layer_ids = [i + self.tap_shift for i in self.target_layer_ids]
+        assert len(set(self.target_layer_ids)) == len(self.target_layer_ids), \
+            "DFlash target_layer_ids must be unique"
         self.block_size = self.read_cfg(int, ["block_size", "dflash_config->block_size"], no_default)
 
         # RoPE
@@ -212,6 +214,13 @@ class DFlashModel(Model):
             "block_table": torch.Tensor
             "cache_seqlens": torch.Tensor
         """
+
+        # Target states arrive in layer execution order. Reorder them only when the checkpoint's
+        # projection expects a different target_layer_ids order.
+        target_layer_ids = self.config.target_layer_ids
+        if target_layer_ids != sorted(target_layer_ids):
+            source_idx = {layer_id: idx for idx, layer_id in enumerate(sorted(target_layer_ids))}
+            target_hidden = [target_hidden[source_idx[layer_id]] for layer_id in target_layer_ids]
 
         # May update a few redundant tokens when batching, but we'd never draft longer than the cache length
         if lengths is not None:
--- a/exllamav3/architecture/dflash2.py
+++ b/exllamav3/architecture/dflash2.py
@@ -1,0 +1,211 @@
+from __future__ import annotations
+from typing_extensions import override
+import torch
+
+from .dflash import DFlashConfig, DFlashModel
+from ..model.config import no_default
+from ..modules.arch_specific.dflash2 import (
+    DFlash2Block, DFlash2DynConv, DFlash2Selector,
+)
+
+# DFlash2 draft model: a DFlash backbone with grouped dynamic convolutions around
+# every attention/MLP sublayer and a top-k candidate selector in place of row-wise argmax.
+#
+# Conventions:
+#   - Block input = [anchor, mask x (block_size-1)]; each remaining row predicts
+#     its own position and the selector walks candidates starting from the anchor.
+#   - Taps: reference reads HF hidden_states[target_layer_ids[i] + 1] = output
+#     of target layer id  =>  tap_shift 0 (exl3 export index = layer output).
+#   - Draft cache ctx K/V come from update_kv_from_target (fc+hidden_norm
+#     projected tap stream), inherited unchanged from the DFlash1 backbone;
+#     per-round writes at the new cache position overwrite the transient
+#     noise-block K/V, reproducing the reference's crop semantics.
+#   - Proposals are the greedy selector path (T-independent), verified by the
+#     stock accept-while-match rule — trivially lossless at every temperature
+#     (per-position output marginal equals the target distribution).
+
+
+class DFlash2Config(DFlashConfig):
+
+    arch_string = "DFlash2DraftModel"
+
+    # Reference extract uses hidden_states[id + 1] == output of layer id
+    tap_shift = 0
+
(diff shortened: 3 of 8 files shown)


turboderp:
There's a couple of things I would change:

- `export_state_order` is not needed. The taps are already exported in execution order through the existing `export_state_layers` set, and `update_kv_from_target` concatenates them before the fc projection. Trying to keep code bloat to a minimum, it would be cleaner to just reorder the list by target_layer_ids in `update_kv_from_target, if `target_layer_ids` isn't always ascending anyway (looks like it from the GLM5.3 DFlash2 model.)

- I don't think there's a reason to override the dynamic draft setting. The existing DFlash loop already runs the drafter at its fixed block size and crops the draft to the given window (which can shrink dynamically if `dynamic_draft` is enabled. What would make more sense is to export a `draft_conf` signal so dynamic drafting is enabled rather than forced-off. Especially for MoE models, the confidence signal is very helpful for truncating the draft and not wasting cycles on validating very unlikely tokens.

- `_DFlash2Norm` is a plain RMS norm. There are already custom kernels for that. It doesn't support bfloat16 outputs, but there's no real reason for bfloat16 anyway. It's a downgrade from the default, in fact, which is to keep the residual in fp32. I would drop the cast on the first line, use the stock RMSNorm module throughout with fp32 input, bf16 weights and fp16 output for the convolution .prepare() that follows (post-norm values are bounded; prefer the higher precision of fp16.) I would run .finish() in fp32 and returns fp32 to add back on the fp32 residual without losing anything.

- `_grouped_dynamic_convolve` is an eager reference implementation, and it's right in the hot path. Really calls for a custom kernel to keep drafting latency to a minimum.

- Changes in attention_fn/torch.py and common.py look like test instrumentation. The spans are redundant with the existing DFlash mechanism, which already sets `causal=False` on the whole call.

I did test it as-is, and on my setup it mostly loses to not using a draft. The accepted draft length is around 2 in most cases, which suggests it could be useful with dynamic draft sizing, but probably isn't otherwise at bsz 1. Some of that could be helped by optimizing away the eager reference code, though. Didn't fully profile everything yet.

RodriMora:
Thanks for the review. I pushed `49293a8`, rebased onto the current `master`.

The update addresses each point from the review:

- Removed `export_state_order`. `update_kv_from_target()` now reorders exported states from execution order into `target_layer_ids` order before the projection.
- Re-enabled dynamic drafting. DFlash2 exports the winning combined selector score, aligned as `[anchor placeholder, proposal scores...]`, through the existing `draft_conf` path.
- Removed `_DFlash2Norm`. The blocks now reuse the stock `RMSNorm`, keep the residual stream and convolution `finish()` in fp32, and use fp16 for post-norm values and convolution `prepare()`.
- Replaced the eager grouped dynamic convolution with a stride-aware Triton kernel. It reads the noncontiguous dynamic-kernel projection views directly, so it does not add a contiguous copy in the hot path. The Torch implementation remains as a fallback.
- Dropped the attention span changes. The attention files now match upstream and DFlash2 uses the existing whole-call `causal=False` behavior.

The focused GPU test script now has 14 tests. It covers tap ordering, confidence alignment, fp32 residual behavior, noncontiguous dynamic views, odd convolution geometry, fp16 output parity, and fp32 values above the fp16 range.

For the checkpoint geometry at batch 1, length 8, hidden size 4096, kernel size 2, and group size 16, the convolution microbenchmark was:

| Path | fp16 prepare | fp32 finish |
| --- | ---: | ---: |
| Triton | 40.6 us | 39.0 us |
| Previous eager implementation | 190.1 us | 214.9 us |

I also ran target-only, fixed seven-token DFlash2, and calibrated dynamic DFlash2 on several 256-token workloads. Each dynamic measurement followed a 256-token same-workload calibration run.

| Workload | Target only | Fixed DFlash2 | Dynamic DFlash2 |
| --- | ---: | ---: | ---: |
| JSON-style structured output | 54.40 tok/s | 70.04 tok/s, +28.7% | 76.68 tok/s, +40.9% |
| Agent-style execution trace | 54.50 tok/s | 30.97 tok/s, -43.2% | 59.59 tok/s, +9.4% |
| Python coding | 54.71 tok/s | 62.48 tok/s, +14.2% | 48.18 tok/s, -11.9% |
| Analytical prose | 54.93 tok/s | 43.08 tok/s, -21.6% | 54.59 tok/s, -0.6% |

The performance result is mixed. Structured output was the strongest case. Dynamic truncation also rescued the agent-style workload from a very poor fixed-window result. It did not help the coding run, and prose was roughly break-even. Across these four equal-length runs, fixed DFlash2 was 14.6% slower in aggregate and dynamic DFlash2 was 6.3% faster.

An earlier general 512-token smoke test was also slower with dynamic DFlash2, 52.0 tok/s versus 56.6 tok/s for target-only, with 2.5 accepted tokens per verification round. That agrees with the original observation that DFlash2 does not win consistently at batch 1.

These are single measured runs, not a broad performance study. This quantized MoE target is not bit-stable between repeated target-only greedy runs, so the different modes can follow different continuations and acceptance profiles. I would not claim a general batch-1 speedup from these numbers. They do show a useful structured-output case, and they show why exporting confidence is important when acceptance is poor.

turboderp:
Okay, I merged this into the `dev` branch but forgot to retarget the PR first, so it'll show as open until `dev` is merged into `master` for the next release.

I had to make a few changes:

- It's now a full model class (deriving from DFlashModel is fine in this case but could become fragile since DFlashModel created a complete model instance before DFlash2Model patched it. Can't guarantee that the two architectures will stay compatible like that.)
- Convolution kernel projections are now also quantized
- Replaced the Triton conv kernel with a faster CUDA kernel (also much less launch overhead)
- Added a custom top-K kernel since Torch was falling back on an extremely slow path and didn't have any efficient options for the sizes involved
- Replaced the selector walk with another custom kernel
- Fixed some blocking H2D and D2H copies

For Qwen3.8-27B, draft overhead is reduced from 3.8 to 2.0 ms.

PR #335, opened 2026-09-04
chat.py: add -gcs to set the generator's max_chunk_size

vcruz305:
`Generator` already accepts `max_chunk_size`, but `chat.py` never passes it, so the example is stuck at the 2048 default with no way to change it short of editing the file.

I hit this while testing long context on a 24 GB card, where prefill chunk size is worth tuning and I wanted to try a few values without keeping a local edit around.

The default here is 2048, matching the `Generator` default, so behavior is unchanged unless you pass the flag.


turboderp:
:+1:

PR #338, opened 2026-09-05
Hoist hot-path scratch allocations in hyperconnection mix and routing

creslinux:
## Problem

Two hot decode paths allocate fresh scratch tensors on every call:

- `HyperConnection._mix` allocates its `dots` / `post` / `mixed` buffers per invocation (2× per block per pass);
- the multi-row router path (`routing_std`) allocates per call.

At decode shapes these are small, frequent allocations that ride the caching allocator's slow path once shapes vary (speculative verify windows change row counts constantly).

## Change

Hoist the scratch into per-instance cached buffers keyed by shape, with a safety note: at most one outstanding result per buffer key, which holds because nothing calls `_mix` on a site instance between its `mix()` and `apply_()` — consumers are stream-ordered before the next same-slot write. No behavior change; the comments mark the invariant for anyone who refactors the site.

## Measurement

+3–6% decode end-to-end on a 4× RTX 3090 layer-split rig (Qwen3.8-Flash-Next EXL3, MTP speculative decoding, measured across repeated matched batteries). Measured before/after on the same prompts with warm-up runs discarded.

### DIFF
--- a/exllamav3/modules/block_sparse_mlp.py
+++ b/exllamav3/modules/block_sparse_mlp.py
@@ -89,9 +89,22 @@ def routing_std(bsz, cfg, y, params):
                 routing_weights *= cfg.per_expert_scale.unsqueeze(0)
             return selected_experts, routing_weights
         else:
-            router_logits = torch.empty((bsz, cfg.num_experts), dtype = torch.half, device = y.device)
-            routing_weights = torch.empty((bsz, cfg.num_experts_per_tok), dtype = torch.half, device = y.device)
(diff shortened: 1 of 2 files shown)


turboderp:
Thanks. I've merged this but also rewritten it a bit since there's already machinery (`g_tensor_cache`) to deal with exactly this. The scratch allocations are now bucketed (slices rather than allocating tensors for each unique batch size) and shared between instances on the same device. Also the MoE allocation was unbounded, meaning each layer created its own set of large, static tensors during prefill.

PR #340, opened 2026-09-06
Generator: Fix multimodal bounds during rewind prefill

BFJ-Concerns:
A banned-string rewind on a recurrent model can replay prefill over generated text. The multimodal mask and MRoPE table only cover the prompt, causing two bounds errors:

- A chunk starting beyond the prompt raises `IndexError` in the `mm_span_prefix` walk.
- A chunk ending beyond the prompt passes an undersized MRoPE table to the unchecked RoPE kernel. This can happen even if the chunk starts inside the prompt.

Guard the mask lookup and extend the table before the model call. Appending text advances the next rotary position and sequence length equally, so `alt_rope_offset` remains valid for decode.

Observed with TabbyAPI on exllamav3 1.4.2, Qwen3.8-27B EXL3 5.0bpw, vision and MTP enabled, on two RTX 3090s. On 1.4.2 the exception disrupted the generator; `dev` already contains prefill exceptions per job but retains both bounds errors.

Nine CPU regression tests exercise the actual `Job.prefill()` and MRoPE builder, intercepting model calls before GPU execution. They cover replay before, at and beyond the prompt, frequency and offset preservation, text-only replay and MTP target inputs. They fail on unpatched code and the mask-only fix, and pass with both fixes. Including the existing related tests: 25 tests and six subtests pass.

```sh
CUDA_VISIBLE_DEVICES='' python -m unittest discover -s tests \
  -p test_multimodal_rewind_prefill.py -v
```

Earlier model runs with both fixes had no crashes or punctuation/script detector hits in 16 image requests across 164 rewinds. That detector is not a general output-quality check. These runs used 1.4.2; CPU checks used its installed extension, whose MRoPE position-builder source matches `dev`. GPU recurrent-state equivalence and Gemma's `atomic_mm_prefill` path remain untested.

### DIFF
--- a/exllamav3/generator/job.py
+++ b/exllamav3/generator/job.py
@@ -1320,11 +1320,21 @@ class Job:
                 # span extends before the chunk, so non-causal attention windows cover the whole
                 # span rather than just the in-chunk suffix
                 mm_span_prefix = 0
-                if self.embeddings:
+                # The mask covers only the prompt; generated tokens are never multimodal
+                if self.embeddings and prefill_start <= len(seq.multimodal_mask):
                     pp = prefill_start
                     while pp > 0 and seq.multimodal_mask[pp - 1]:
                         mm_span_prefix += 1
                         pp -= 1
+
+                # Rewind prefill can exceed the prompt-length table, which the RoPE kernel reads unchecked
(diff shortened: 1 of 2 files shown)


turboderp:
Thanks. :+1:

Turns out it also breaks Gemma4 which has non-causal image token spans. I'll fix that separately.

PR #341, opened 2026-09-06
MoE CPU offload: zero-copy streamed prefill from a page-locked shared arena (~2x long-context prefill)

DenysAshikhin:
## Purpose

GPU-streamed prefill for CPU-offloaded MoE experts currently goes through a host staging copy:
a stager thread memcpy's each expert's packed weights from the worker's arena into a pinned
staging ring, and only then DMAs them to VRAM. On the tested machine that memcpy step measured
only 11-18 GB/s and was the bottleneck for long prefill.

This patch removes the staging copy. The CPU worker's expert arena becomes a set of 1 GiB
shared-memory chunks that the parent maps and `cudaHostRegister`s at worker start. Streamed
prefill then DMAs `[gate|up|down]` expert blocks straight from the arena into raw VRAM slots
and un-swizzles/repacks them on the GPU. Every streamed byte costs one DRAM read instead of a
memcpy plus a DMA, and the handoff becomes PCIe-bound (26-30 ms per ~1 GiB layer on a 4090).

## Changes

- Remove the stager thread, stage-job ring and pinned staging slots
  (`moe_handoff.cu/.h`, `moe_mul1.cpp/.h`). `EXL3_MOE_CPU_STAGE_THREADS` is gone.
- `moe_cpu_host.py`: register the shared arena for DMA, stream directly from it, repack on the
  GPU via `moe_unswizzle_trellis`. Two raw DMA landing slots are added alongside the existing
  VRAM ring; VRAM cost is `(WSLOTS + 2) * WSLOT_MB`.
- gate/up and down GEMV phases partition work in 8-tile band units with one contiguous block
  per worker. Bit-exact with the previous assignment.
- `EXL3_MOE_MEMOPS` now defaults to `0` on Windows: under WDDM the memop handshake measured
  slower than the kernel waits (23-26 vs 28-30 tok/s single-token decode).
- `EXL3_MOE_STREAM_T` default scale base lowered from 16 to 8.
- New `EXL3_MOE_STREAM_PROF` per-layer streamed-prefill profiler (diagnostic only).
- Bandwidth probe warms the link for 250 ms and takes best-of-8.
- `tests/test_moe_cpu_offload.py`: partition invariance, swizzled == native, un-swizzle exactness.
- `doc/env_vars.md` updated accordingly.

## Results

RTX 4090, Ryzen 9 7900X, `td_flash-next_4.05bpw_h6_ng6`, `perf.py -mcs 410 -mct 12 -cs 32768
-chunk_size 4096 -ngr -max_length 32768`, `EXL3_LOAD_ARENA=1 EXL3_MOE_MEMOPS=0`,
`PYTORCH_ALLOC_CONF=backend:native`, torch 2.13.0+cu132. This branch: 5 runs per platform,
median (min-max). Upstream `c93f3c6`: same command, 1-2 runs, observed values.

| Prefill tok/s | 4096 | 8192 | 16384 | 32768 |
|---|---:|---:|---:|---:|
| Upstream, Windows | 895 | 855-1025 | 836 | 824-1020 |
| Upstream, WSL2 | 1131 | 931-1039 | 906 | 895-1045 |
| This branch, Windows (`/Ox`) | 1944 (1675-2128) | 1814 (1592-1948) | 1794 (1622-2009) | 1754 (1588-1980) |
| This branch, WSL2 (Ubuntu 24.04, `-Ofast`) | 2044 (1689-2133) | 1928 (1612-2003) | 1962 (1642-2038) | 1927 (1614-2016) |

| Decode tok/s, median | ctx 0 | ctx 4096 | ctx 32512 |
|---|---:|---:|---:|
| Upstream, Windows | 24.2 | 21.9 | 25.0 |
| Upstream, WSL2 | 25.8 | 25.1 | 26.4 |
| This branch, Windows | 33.6 | 33.4 | 33.0 |
| This branch, WSL2 | 28.9 | 29.4 | 29.7 |

Decode varies with machine state more than with the patch (one Windows run sat at 24-27
across all contexts); the ctx-2048 placement sweep is excluded above.

## Note on `-Ofast` (pre-existing, not changed here)

`test_moe_cpu_offload.py` originally used unit-scale hidden states, which drive the gate
activation past `exp()` overflow. The `/Ox` Windows build recovers; the `-Ofast` Linux
build (setup.py, non-Windows) yields NaN rows. Upstream's unmodified kernel does the same on
those inputs, and an `-O3` build is bit-identical to Windows. The test now uses a realistic
activation scale. Worth a separate look at the fast-math flag upstream.


treo:
I ran some tests on my system.

```
OS: Fedora Linux 44 (KDE Plasma Desktop Edition) x86_64
Kernel: Linux 7.1.13-200.fc44.x86_64
CPU: AMD Ryzen 9 5950X
GPU: NVIDIA GeForce RTX 3090 
RAM: 128 GB DDR4 @ 3200 MT/s
```

The performance uplift looks pretty good.

### 2.05bpw Qwen 3.8 Flash Next, Full Context (256K), CPU Moe Split 380
```
python ../exllamav3/eval/perf.py --model_dir ../../tabbyAPI/models/Qwen3.8-Flash-Next-exl3-2.05bpw_h4_ng4/ --moe_cpu_split 380 --max_length 262144 -cs 262144 --chunk_size 2048 -ngr
```
Before (See #331):
image

After:
```
 -- Bitrate: 2.24 bpw / 4.01 bpw (head)
 -- Chunk size: 2048

Prefill:
Length      256:     357.80 tokens/s
Length      512:     665.16 tokens/s
Length     1024:    1087.85 tokens/s
Length     2048:    1373.94 tokens/s
Length     4096:    1354.71 tokens/s
Length     8192:    1333.69 tokens/s
Length    16384:    1321.35 tokens/s
Length    32768:    1314.82 tokens/s
Length    65536:    1328.04 tokens/s
Length   131072:    1323.03 tokens/s
Length   262144:    1292.55 tokens/s

Generation
Context       0: S=1      38.86 tokens/s
Context     256: S=1      38.76 tokens/s
Context     512: S=1      38.39 tokens/s
Context    1024: S=1      35.98 tokens/s
Context    2048: S=1      37.17 tokens/s
Context    4096: S=1      37.78 tokens/s
Context    8192: S=1      38.43 tokens/s
Context   16384: S=1      38.66 tokens/s
Context   32768: S=1      37.62 tokens/s
Context   65536: S=1      36.63 tokens/s
Context   31072: S=1      39.46 tokens/s
Context   61888: S=1      38.36 tokens/s
```

### 4.05bpw Qwen 3.8 Flash Next, Full Context (256K), CPU Moe Split 448
```
python ../exl3-zero-copy/eval/perf.py --model_dir ../../tabbyAPI/models/Qwen3.8-Flash-Next-exl3-4.05bpw_h6_ng6/ -mcs 480 --max_length 262144 -cs 262144 --chunk_size 2048 -ngr 
```

**This crashes on load**. I can see that systemd is collecting a coredump, so there must be a segmentation fault. From the Python side it just shows an `EOFError` from the multiprocessing package.

Reducing the max length by just 256, seems to sidestep the error (but triggers a bug in the perf.py script because it isn't doubling up to the end).

```
python ../exl3-zero-copy/eval/perf.py --model_dir ../../tabbyAPI/models/Qwen3.8-Flash-Next-exl3-4.05bpw_h6_ng6/ -mcs 448 --max_length 261888 -cs 261888 --chunk_size 2048 -ngr  -spf
```

Before:
image


After:
```
 -- Bitrate: 4.42 bpw / 6.01 bpw (head)
 -- Chunk size: 2048

Prefill:
Length      256:     183.22 tokens/s
Length      512:     351.63 tokens/s
Length     1024:     614.44 tokens/s
Length     2048:     912.11 tokens/s
Length     4096:     872.14 tokens/s
Length     8192:     841.99 tokens/s
Length    16384:     838.16 tokens/s
Length    32768:     844.70 tokens/s
Length    65536:     867.90 tokens/s
Length   131072:     871.52 tokens/s

Generation
Context       0: S=1      19.62 tokens/s
Context     256: S=1      21.53 tokens/s
Context     512: S=1      21.68 tokens/s
Context    1024: S=1      21.82 tokens/s
Context    2048: S=1      21.83 tokens/s
Context    4096: S=1      19.66 tokens/s
Context    8192: S=1      22.30 tokens/s
Context   16384: S=1      22.10 tokens/s
Context   32768: S=1      22.57 tokens/s
Context   65536: S=1      22.09 tokens/s
Context  131072: S=1      19.49 tokens/s
Context  261632: S=1      22.01 tokens/s
```

---
## Edit

I've found the reason for the crash. And it was a bit of a wild chase.

Fedora by default sets /dev/shm to be only 50% of the available memory. After changing that to allow full usage, it still failed though. And apparently Fedora *also* sets a per user quota of 40% on that mount. And no matter what I tried I couldn't actually change that yet.

But If I repeat the run as root, it doesn't get limited like that and I can finish a benchmark run, that looks essentially the save as above.

As that is quite a foot gun, it would be nice to have an explicit check for it.

Edit 2: After setting the /dev/shm size using /etc/fstab and rebooting, the automatically set quota also increased.

According to https://lwn.net/Articles/1028275/#:~:text=Quotas,shm%2E
> Quotas for tmpfs
> 
> With systemd v258, users are automatically assigned a quota for [/dev/shm](https://man7.org/linux/man-pages/man7/shm_overview.7.html) and /tmp if /tmp is backed by a tmpfs. The default is 80% of the size assigned to the tmpfs. The goal for this feature is to prevent any user from effectively performing a denial-of-service attack by filling up /tmp or /dev/shm.

So what I was seeing 40% because I had to increase the size of it to 128GB. 

My remark about SHM being a bit of a foot-gun on Linux remain. It needs checking before trying to allocate, and a SIGBUS handler while allocating, so the user doesn't have to go on a wild goose chase trying to figure out why things crash.

It would probably also make sense to document how the user can increase the 80% limit, as a user with a 3090 and 64 GB RAM should still have just enough memory to run the 4 bit quant, but the limit would stop them from doing so.

treo:
I've messed around with an alternative to to using Python's Shared Memory setup, by using cmemfd_create. It seems to work exactly as fast, without creating any /dev/shm files, and is therefore not subject to those quota rules. 

But it has another significant drawback: It is only available if the python version was built in an environment where it was available. There are ways to get around that, and, in fact, I had to use that workaround while trying it, but they are pretty ugly.

The 2x prefill increase is certainly worth trying to find a solution.

DenysAshikhin:
> I've messed around with an alternative to to using Python's Shared Memory setup, by using cmemfd_create. It seems to work exactly as fast, without creating any /dev/shm files, and is therefore not subject to those quota rules.
> 
> But it has another significant drawback: It is only available if the python version was built in an environment where it was available. There are ways to get around that, and, in fact, I had to use that workaround while trying it, but they are pretty ugly.
> 
> The 2x prefill increase is certainly worth trying to find a solution.

Can you go a bit more detail - I want to look into it. Because I am looking for other ways to hit these numbers which are hitting limited improvements. It's hard to beat the simplicity of just less work read/write being done in the first place

treo:
[0001-memfd-arena.patch](https://github.com/user-attachments/files/31942852/0001-memfd-arena.patch)
[0002-memfd-resolver.patch](https://github.com/user-attachments/files/31942856/0002-memfd-resolver.patch)

Here you have the full detail :) The first patch changes your implementation to support `os.memfd_create`, so it does 3 things: /dev/shm, memfd_create and huge pages.

The second patch is all about making use of memfd_create even when it is not available in the os package. So, the workaround that I've been talking about.

In my benchmarking the huge pages support didn't really pay off. I've tried it both with 2m and 1g page size, and both prefill and decode were essentially the same.

Ph0rk0z:
Damn.. that's a long ass patch. I just got around conda not having it by using:
```


import ctypes, os

if not hasattr(os, "memfd_create"):
    libc = ctypes.CDLL(None, use_errno=True)
    libc.syscall.restype = ctypes.c_long

    # syscall number for memfd_create
    _SYS_memfd_create = 319 if os.uname().machine == "x86_64" else 279  # aarch64

    def memfd_create(name: str, flags: int = 0) -> int:
        fd = libc.syscall(_SYS_memfd_create, os.fsencode(name), flags)
        if fd < 0:
            err = ctypes.get_errno()
            raise OSError(err, os.strerror(err))
        return fd

    os.memfd_create = memfd_create
```

Is this one faster than the version currently in dev? I don't think I got much out of that one.

DenysAshikhin:
> Damn.. that's a long ass patch. I just got around conda not having it by using:
> 
> ```
> 
> 
> import ctypes, os
> 
> if not hasattr(os, "memfd_create"):
>     libc = ctypes.CDLL(None, use_errno=True)
>     libc.syscall.restype = ctypes.c_long
> 
>     # syscall number for memfd_create
>     _SYS_memfd_create = 319 if os.uname().machine == "x86_64" else 279  # aarch64
> 
>     def memfd_create(name: str, flags: int = 0) -> int:
>         fd = libc.syscall(_SYS_memfd_create, os.fsencode(name), flags)
>         if fd < 0:
>             err = ctypes.get_errno()
>             raise OSError(err, os.strerror(err))
>         return fd
> 
>     os.memfd_create = memfd_create
> ```
> 
> Is this one faster than the version currently in dev? I don't think I got much out of that one.

I am working on making a dev merged one that is merge friendly

Ph0rk0z:
I cloned the branch and left pinned arena, didn't see any speedup but only slowdown. But memory use seemed lower?

Do I need to use some of this?

EXL3_LOAD_ARENA=1 EXL3_MOE_MEMOPS=0,
PYTORCH_ALLOC_CONF=backend:native

This isn't windows only?

DenysAshikhin:
> I cloned the branch and left pinned arena, didn't see any speedup but only slowdown. But memory use seemed lower?
> 
> Do I need to use some of this?
> 
> EXL3_LOAD_ARENA=1 EXL3_MOE_MEMOPS=0, PYTORCH_ALLOC_CONF=backend:native
> 
> This isn't windows only?

wdym?
And what machine/model are you running? And what  chunk size?

Reason being Turbo took the core idea and implemented it on main already, this pr has been reworked now to just make it work on Windows.

The numbers from initial post were before Turbo implmented core idea in main exl3.

Ph0rk0z:
I thought this was the improved version but I guess nm.

DenysAshikhin:
> I thought this was the improved version but I guess nm.

Nah, it was, then Turbo implemented it in a more streamlined way so this PR got stranded. Now its just for windows to close it out

PR #346, opened 2026-09-08
Pinned vision: rebuild MLP native handles after pin_linears (VRAM doublecount)

DenysAshikhin:
TL:DR
We are holding onto more VRAM than we need to.

`MLP.load` builds `BC_MLP` from the linears' native handles before the loader calls
`pin_linears`. Pinning swaps the linears' handles, but `BC_MLP` keeps the old ones and the
original VRAM trellis tensors with them, so a pinned vision tower holds its weights twice.
`GatedMLP` additionally keeps a `MultiLinear` pointer table of freed device addresses that
the prefill path still reads.

Override `pin_linears` on both classes to pin the children and rerun `load_local`.

Qwen3-VL tower, RTX 4090: pinned-load VRAM 212.7 MiB -> 19.1 MiB, embeddings unchanged.
Adds a GPU test gated on `EXL3_TEST_VISION_MODEL`.

### DIFF
--- a/exllamav3/modules/mlp.py
+++ b/exllamav3/modules/mlp.py
@@ -237,6 +237,16 @@ class MLP(Module):
         self.bsz1_pa_args = []
 
 
+    @override
+    def pin_linears(self):
+        # Pinning swaps the children's native linear handles; the graphed bsz-1 path built by
+        # load_local captured the old ones and, through them, the original VRAM trellis. Rebuild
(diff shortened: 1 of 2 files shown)


Scorpoon:
@DenysAshikhin Applied patch on my system, saved me 380MB on Qwen 3.8 Flash-Next. No errors so far. Test looks good to me :+1:

turboderp:
:cat2:

PR #349, opened 2026-09-08
CPU MoE: AVX-512BW kernel tier, plus three kernel/scheduling improvements

Lowkey-Loki-SN:
### Why

`cpu/moe_mul1.cpp` dispatches on `scalar / avx2 / avx512-vnni / avx512-vbmi`, and the 512-bit paths require VNNI. Skylake-SP/X class CPUs (1st-gen Xeon Scalable, Xeon W-21xx/31xx, Core-X) have AVX-512F/BW/VL but no VNNI, so they ran the 256-bit AVX2 tier. The only VNNI dependency in the 512-bit kernel is the accumulate (`vpdpbusd`); the extraction is pure AVX-512F, and the AVX2 tier already has the VNNI-free, bit-exact substitute.

### Commits

1. **`Isa::Bw` tier.** The VNNI kernel's dword extraction and band structure with the AVX2 tier's `vpmaddubsw`/`vpmaddwd` accumulate. Separate functions because a function carries one target attribute and compiling this accumulate under the VNNI target would permit contracting it into `vpdpwssd`. Exposes `exl3_moe_cpu_has_avx512_bw()`, accepts `EXL3_MOE_CPU_MAX_ISA=bw`, enables the swizzled layout for the tier, updates `doc/env_vars.md`.
2. **Even flat tile partition in the few-GEMV regime** (all tiers). Whole-GEMV assignment left a 2:1 worker imbalance whenever `2 × cold experts` wasn't a multiple of the worker count (16 GEMVs on 20 workers: twelve single-worker GEMVs set the phase time). For ≤ 4×workers GEMVs the phase's tiles are one flat range split evenly in 8-tile groups; above that the strided whole-GEMV scheme stays so L3 keeps serving repeated expert chunks during prefill. `gemv_assignment`/`tile_split` fold into `assign_gemvs`.
3. **Per-lane variable-shift funnel merge** in `dword_codes` (BW and VNNI tiers): 8 → 4 uops per k-row per tile, and GCC fuses the trailing or+and into `vpternlogd`. Bit-identical codes.
4. **Force-inline the BW row chain.** GCC outlined it behind a call per tile step, and with all zmm registers caller-saved that reloaded every live constant.
5. **`tests/test_moe_cpu_tiers_.py`.** Every tier the CPU supports, one subprocess each via `EXL3_MOE_CPU_MAX_ISA`, checked against the AVX2 tier (int8 tiers to float rounding, scalar fp32 reference to ~1%) over K1–8, gated/gateless, 1–4 rows per chunk, the strided regime, and the swizzled layout on the tiers that consume it. **On a VBMI machine this runs vbmi, vnni, bw, avx2 and scalar in one go**, which is how the BW tier can be validated without the hardware.

### Numbers

Xeon Gold 6148 (20C, AVX-512F/BW/DQ/VL, no VNNI), GCC 13, cold experts streamed from DRAM, 20 threads, median GB/s of trellis bytes, single-token decode, 8 cold experts per layer:

| shape | avx2 (before) | 1: bw | 2: +partition | 3: +varshift | 4: +inline |
|---|---|---|---|---|---|
| 4096×2048 K3 | 15.6 | 24.0 | 30.6 | 38.3 | 41.4 |
| 4096×2048 K4 | 21.1 | 35.4 | 39.2 | 51.8 | 57.0 |
| 2560×640 K4 | 17–25 | 37.6 | ≈ | 44.5 | 46.3 |

Partition by cold experts per layer (K3): 5 neutral, 6 +40%, 7 +33%, 8 +17%; 10 (already even) neutral. Batched rows (4 per expert): 11.6 → 20.0. 10 → 20 threads still scales 1.73×, so the kernel remains compute-bound here.

End to end (TabbyAPI): GLM-5.3-Flash 3.05bpw, 232/288 experts per layer on CPU: **4.3–4.9 → 11.7 T/s** (8k-token generation). Qwen3.8-Flash-Next 4.05bpw, 340/512 on CPU: **24.6 → 31.8 T/s** after commits.

### Correctness

Strict-FP builds (`-O3 -fno-fast-math -ffp-contract=off`): bw ≡ avx2 bit-for-bit over 640 cases (K1–8, 1–5 tokens, gated/gateless, 1 and 20 threads, partial bands, swizzled layout) plus 256 many-chunk cases for the strided regime; avx2 and scalar tiers unchanged bit-for-bit through all four commits. Under `-Ofast` tiers differ only by epilogue reassociation (≤ 1e-4 rel). Disassembly: no VNNI/VBMI instructions in `bw_*`.

### Caveats

- Not runtime-tested on VNNI/VBMI hardware. Commit 1's change there is attribute-only (helpers still inline, verified in the disassembly). Commit 3 changes VNNI-tier codegen (`vpsrlvd/vpsllvd` for immediate shifts + blend); expected neutral-to-positive (single-uop, latency-1 on Skylake-SP through Sapphire Rapids and Zen 4/5), a Cascade Lake A/B would settle it. The VBMI tier only meets this code on its K8 route.
- Swizzled layout for the BW tier rests on one machine's measurements (+2–29%); `EXL3_MOE_CPU_SWIZZLE=0` opts out.
- The Windows branch of `detect_isa()` and `M1_ALWAYS_INLINE` (`__forceinline`) are compile-untested here, like the file's existing Windows-specific code.

### Disclaimer

Implementation code fully written with AI-assistance but was manually tested and validated on my setup. The CPU-offloaded gains are instantly noticeable but remain untested on other CPU platforms. For now, I'll mark this PR as draft until maintainers chip in and deem this ready,

IF bits and pieces of this are to be taken with the decision to write a clean PR from scratch, that's understandable and you have my full approval. Feel free to!

### DIFF
--- a/doc/env_vars.md
+++ b/doc/env_vars.md
@@ -268,13 +268,17 @@ only a small amount of shared-memory overprovisioning, not runtime.
 
 ### `EXL3_MOE_CPU_MAX_ISA` (default: unset, auto-detect)
 
-Caps the CPU kernel's runtime ISA detection at `scalar`, `avx2`, `vnni`/`avx512`, or `vbmi`,
-for testing a lower-tier kernel path on hardware that supports better. The `vbmi` tier
+Caps the CPU kernel's runtime ISA detection at `scalar`, `avx2`, `bw`/`avx512bw`,
+`vnni`/`avx512`, or `vbmi`, for testing a lower-tier kernel path on hardware that supports
+better. The `bw` tier covers AVX-512F/BW/VL hardware without VNNI (Skylake-SP/X: 1st-gen Xeon
+Scalable, Core-X), which previously fell through to `avx2`: the `vnni` dword kernel with the
+AVX2 tier's vpmaddubsw/vpmaddwd accumulate, ~1.5x the `avx2` tier's cold-expert decode
+throughput on a Xeon Gold 6148. The `vbmi` tier
 (AVX512-VBMI byte-gather state extraction, Zen 4+ / Ice Lake+; Cascade/Cooper Lake have VNNI
 without VBMI and stay on the `vnni` tier) is 15-70% faster than the dword scheme depending on
 bitrate. Never upgrades past what the CPU actually supports; unrecognized values are ignored.
 Read once per process (parent and worker independently), so it must be set before either is
-started. Note that capping below `vbmi` also disables the swizzled weight layout (see
+started. Note that capping below `bw` also disables the swizzled weight layout (see
 `EXL3_MOE_CPU_SWIZZLE`).
 
 ### `EXL3_MOE_CPU_SWIZZLE` (default: `1`)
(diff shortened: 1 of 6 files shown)


turboderp:
This is pretty great. :cat2: 

I'm also seeing a 10% improvement on VBMI here, and a 42% improvement on VNNI with all changed applied including the swizzle that otherwise wasn't beneficial on VNNI. And the two tiers are now at parity on my Threadripper 7960X. AVX2 is up by 16%, and BW works here although it sits about 9% below VBMI and VNNI.

And Skylake, Xeon-W and Core-X should benefit from the BW tier. Good stuff.

I changed the gating a bit and updated the docs (swizzling is applied by default on all AVX512 tiers now.)

Lowkey-Loki-SN:
Glad to hear. With this landed, exllamav3 is now the fastest engine on my setup for running CPU-offloaded models. Thank you for this wonderful project!

PR #354, opened 2026-09-10
Port the p2b cooperative fused-MoE kernel from vllm-exl3 (opt-in, +25-76% decode on layer split)

creslinux:
Ports the p2b cooperative fused-MoE kernel into exllamav3 and routes decode-shaped
`BlockSparseMLP` passes through it behind `EXL3_P2B_MOE`, default off.

## Credit first

**The kernel is not mine.** It was written by @vcruz305 for
[vllm-exl3](https://github.com/vcruz305/vllm-exl3) (MIT), where it was already measured
against exllamav3's own MoE path. This PR ports it into this tree; the file header carries
the original attribution and licence. What is mine is the port, the integration, the
debugging and the numbers below.

## What the port changed

- Extended from one input row per launch to a full row-expert slot table, so a whole verify
  window (`bsz * top_k` slots) runs as one cooperative launch instead of one per row.
  Zero-weight slots are skipped at all eight per-slot work sites.
- Dimensions became runtime parameters. The reference build hardcoded hidden 4096 /
  intermediate 2048; they now need only be multiples of 128 for the Hadamard warp tiling.
- Fixed the codebook dispatch. The reference selected the codebook with a truthiness test on
  a value that is never zero, so every launch decoded with the wrong table and `mul1` was
  never bound. The tell was two supposedly different codebooks producing a bit-identical
  error to five digits.

## Measured

`EXL3_P2B_MOE` 0 vs 1 on this branch, nothing else changed. 4x RTX 3090, layer split,
Qwen3.8-Flash-Next EXL3 4.05bpw, MTP drafting, Q6 KV, ~1200-token natural-length
generations, both shapes warmed twice, four runs per arm:

| | off | on | gain |
|---|---:|---:|---:|
| code | 57.4 tok/s | **100.9** | +76% |
| prose | 66.4 tok/s | **82.7** | +25% |

Runs, off: code 62.5 / 55.4 / 59.4 / 49.6, prose 45.1 / 66.7 / 66.0 / 69.1.
Runs, on: code 99.2 / 103.9 / 102.3 / 99.5, prose 86.5 / 82.5 / 81.9 / 83.0.

The on arm is tight (under 5% spread), the off arm is not, so read the medians. Every off
run is below every on run in both shapes.

## Where the gain actually comes from

Worth being straight about this: under layer split the kernel's headline mechanism, skipping
zero-weight slots, **cannot** be responsible, because every expert is local and no slot is
masked. The gain here is the integration work in the second commit — staged scratch replacing
seven per-call allocations plus a host memset per layer per pass, the accumulator returned as
fp32 instead of round-tripping through half, the local-id mapping collapsed from ~9
elementwise launches into one kernel, and the shared expert through its own multi-row BC graph
rather than the eager tail. On an expert-range shard, where most slots are masked, the
compaction does the work instead.

## Risk

- Additive only: 730 insertions, no existing line modified.
- `EXL3_P2B_MOE` defaults to off, so nothing changes unless it is set.
- No architecture-specific code: plain `__shfl_sync` and cooperative groups, no tensor-core
  intrinsics, no `__CUDA_ARCH__` branches. Grid size is occupancy-derived.
- Requires cooperative launch (`cudaLaunchCooperativeKernel`) and grid-wide sync, so
  occupancy matters and the gain will vary with SM count. **Tested on Ampere (sm_86) only.**
  ROCm is untested and `cooperative_groups` grid sync maps differently on HIP.
- Constraints inherited from the kernel: MCG codebook only, K in {2, 3, 4}, uniform across
  gate/up/down.

The launch is wrapped in the layer's device context. Without that, the layer-split forward
leaves the current CUDA device at `cuda:0` process-wide, which reaches `cuda:1` tensors
through peer access but faults with an illegal access on the first non-peer pair — any model
spanning three or more cards. The error surfaces asynchronously at whatever kernel checks
next, which is why it was originally misattributed to autotune.

## Diagnostic

`p2b_stage_debug` runs the kernel to phase N of four and returns the staged intermediates, so
a parity failure can be bisected to a phase rather than a tile. It is how the codebook fault
above was found. Every production call site runs all four phases, so it costs nothing at
inference. Documented at the kernel parameter, the header declaration and the binding; happy
to drop it if you would rather not carry it.

### DIFF
--- a/exllamav3/exllamav3_ext/bindings.cpp
+++ b/exllamav3/exllamav3_ext/bindings.cpp
@@ -31,6 +31,8 @@
 #include "quant/exl3_devctx.cuh"
 #include "quant/exl3_moe.cuh"
 
+#include "quant/p2b_moe.cuh"
+
 #include "generator/strings.h"
 #include "generator/sampling_basic.cuh"
 #include "generator/sampling_extra.cuh"
@@ -144,6 +146,8 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
     m.def("reconstruct_had_slice", &reconstruct_had_slice, "reconstruct_had_slice");
     m.def("reconstruct_slice", &reconstruct_slice, "reconstruct_slice");
     m.def("had_r_128", &had_r_128, "had_r_128");
+    m.def("p2b_fused_moe", &p2b_fused_moe_cuda, "p2b_fused_moe",
+          py::call_guard<py::gil_scoped_release>());
     m.def("exl3_gemm", &exl3_gemm, "exl3_gemm");
     m.def("exl3_gemv", &exl3_gemv, "exl3_gemv");
     m.def("exl3_gemm_num_kernel_shapes", &exl3_gemm_num_kernel_shapes, "exl3_gemm_num_kernel_shapes");
--- a/exllamav3/exllamav3_ext/bindings.cpp
+++ b/exllamav3/exllamav3_ext/bindings.cpp
@@ -148,6 +148,12 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
     m.def("had_r_128", &had_r_128, "had_r_128");
     m.def("p2b_fused_moe", &p2b_fused_moe_cuda, "p2b_fused_moe",
           py::call_guard<py::gil_scoped_release>());
+    m.def("p2b_map_slots", &p2b_map_slots, "p2b_map_slots",
+          py::call_guard<py::gil_scoped_release>());
+    m.def("p2b_stage_debug", &p2b_stage_debug_cuda,
+          "p2b_stage_debug: run the p2b MoE kernel to phase N and return the staged "
+          "intermediates (diagnostic; not used at inference)",
+          py::call_guard<py::gil_scoped_release>());
     m.def("exl3_gemm", &exl3_gemm, "exl3_gemm");
     m.def("exl3_gemv", &exl3_gemv, "exl3_gemv");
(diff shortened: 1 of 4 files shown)


vcruz305:
Hey Creslinux, thank you so much for the recognition, as a contributor to exllamav3 I was already working on submitting my work in a PR under my username and would greatly appreciate if you can rescind this and allow me to submit the PR on my end! 🫡 Thank you!

creslinux:
Of course, and no offence intended by the timing. Closing this now so you can submit under your own name.

Everything from the port is on `creslinux/exllamav3` branch `pr/p2b-moe`, two commits on current `dev`, in case any of it saves you time. Take whatever is useful:

- the slot-table extension (whole verify window in one cooperative launch) and runtime dimensions instead of the hardcoded 4096/2048;
- a codebook dispatch fix: the reference selects the codebook with a truthiness test on a value that is never zero, so every launch decodes with the wrong table and `mul1` is never bound. The tell was two different codebooks producing a bit-identical error to five digits;
- the integration follow-ons that carry the layer-split gain (staged scratch, fp32 accumulator returned directly, one-kernel slot map, shared expert through its own BC graph), and the device guard, without which the first non-peer card pair faults inside the kernel;
- `p2b_stage_debug`, which runs the kernel to phase N and returns the intermediates. It is how the dispatch fault was found.

Measured on 4x3090 layer split, Qwen3.8-Flash-Next 4.05bpw with MTP: +18-23% decode against our own tree, +25% prose against plain dev. I would treat the +76% code figure in this PR's description with caution; its control arm was noisier than it should have been and I was about to re-measure it.

Happy to test your PR on this rig when it is up.

PR #356, opened 2026-09-10
Add the p2b cooperative fused-MoE kernel from vllm-exl3 (opt-in via EXL3_P2B_MOE)

vcruz305:
Brings the p2b cooperative fused-MoE kernel from the [vllm-exl3](https://github.com/vcruz305/vllm-exl3) plugin (MIT) into exllamav3, behind `EXL3_P2B_MOE`. Default off, so nothing changes unless it is set.

### Credit

The port is @creslinux's work. He opened it as #354, and closed it so it could go up from here; the kernel is mine but the exllamav3 port, the integration follow-ons and the `p2b_stage_debug` tool are his, and both commits carry him as co-author. He also raised the caveat on the numbers below, which is his own, not something I found.

### What it does

Decode-shaped passes (`bsz <= MAX_BSZN`) route through `ext.p2b_fused_moe` instead of the dense BC graph. Four grid-synchronised phases over the active slots: input Hadamard, batched gate/up GEMV through the existing exl3 GEMV path, SwiGLU plus down-input Hadamard, batched down GEMV, then output Hadamard and weighted atomic accumulation.

Grid size is occupancy-derived and every phase grid-strides over its work items, so there is no architecture-specific code: plain warp shuffles and cooperative groups, no tensor-core intrinsics, no compute-capability branches.

The kernel takes an explicit row-expert slot table rather than the dense `(row x top_k)` form, so the caller maps global expert picks to local ids and clamps out-of-range slots to expert 0 with zero weight. Under an expert-range shard a rank then pays only its active slots. No argsort, no host sync, shapes stay fixed.

### Changes against the vllm-exl3 original

- One input row per launch becomes a full row-expert slot table, so a whole verify window runs as one cooperative launch instead of one launch per row.
- Dimensions are runtime parameters. The original hardcoded hidden 4096 and intermediate 1024/2048; both now need only be multiples of 128 for the Hadamard warp tiling.
- Codebook selection is generalised. The original is MCG-only by construction, fixing the codebook template argument at 1 and rejecting anything else. It is now a template parameter chosen the way `exl3_gemv.cu` chooses it (0 default, 1 MCG, 2 mul1), with both flags at once rejected. Nine instantiations, 3 K by 3 codebook.
- The fp32 accumulator is returned directly rather than rounded to half and converted back in Python.
- `p2b_stage_debug` runs the kernel to phase N and returns the intermediates, so a parity failure bisects to a phase rather than a tile.

Constraints: K in {2, 3, 4}, uniform across gate/up/down, matching the MultiLinear pointer tables.

### Measurements

These are @creslinux's, on 4x3090 layer split, Qwen3.8-Flash-Next 4.05bpw with MTP: **+18-23% decode** against his own tree, **+25% prose** against plain `dev`.

He asked that the +76% figure from #354's description be disregarded: its control arm was noisier than it should have been. I have not reproduced either number on my own hardware, so please read them as his measurements rather than as a verified claim of this PR.

### What I verified

- Compiles clean for `sm_121a` under CUDA 13.0, all nine template instantiations, no warnings referencing `p2b_moe.cu`.
- Rebased onto current `dev` (`8473340`). Upstream has only touched `coop_autotune.cu` since the port was written, which this does not.

### What I changed on top of the port

- Scratch is sized from `MAX_BSZN` rather than a separate hardcoded 16. `bszn_eligible` already gates the path at `bsz <= MAX_BSZN` (8), so the buffers were double what anything could reach, and they can no longer silently diverge from the bound.
- `unload()` releases the p2b buffers through a `_p2b_reset()` helper. They are per-module statics pinning device memory for the life of the module and bound to the device it loaded on, so a module reloaded onto another device would otherwise keep launching against the old one.
- The output parameter is typed `float*` to match what the launcher passes. It was declared `half*` and cast at its single use, which compiled only because `cudaLaunchCooperativeKernel` passes arguments through an untyped `void*` array.
- Normalised a `__restrict__` qualifier that was attached to the pointee rather than the pointer, unlike its nine sibling parameters.
- The engagement notice prints once per process instead of once per layer.

### Not covered

Correctness parity against the dense path on this hardware, and any measurement of my own. Happy to add either if you want them before this is considered. @creslinux offered to test on the 4x3090 rig once it is up.

### DIFF
--- a/exllamav3/exllamav3_ext/bindings.cpp
+++ b/exllamav3/exllamav3_ext/bindings.cpp
@@ -31,6 +31,8 @@
 #include "quant/exl3_devctx.cuh"
 #include "quant/exl3_moe.cuh"
 
+#include "quant/p2b_moe.cuh"
+
 #include "generator/strings.h"
 #include "generator/sampling_basic.cuh"
 #include "generator/sampling_extra.cuh"
@@ -144,6 +146,14 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
     m.def("reconstruct_had_slice", &reconstruct_had_slice, "reconstruct_had_slice");
     m.def("reconstruct_slice", &reconstruct_slice, "reconstruct_slice");
     m.def("had_r_128", &had_r_128, "had_r_128");
+    m.def("p2b_fused_moe", &p2b_fused_moe_cuda, "p2b_fused_moe",
+          py::call_guard<py::gil_scoped_release>());
+    m.def("p2b_map_slots", &p2b_map_slots, "p2b_map_slots",
+          py::call_guard<py::gil_scoped_release>());
+    m.def("p2b_stage_debug", &p2b_stage_debug_cuda,
+          "p2b_stage_debug: run the p2b MoE kernel to phase N and return the staged "
+          "intermediates (diagnostic; not used at inference)",
+          py::call_guard<py::gil_scoped_release>());
     m.def("exl3_gemm", &exl3_gemm, "exl3_gemm");
     m.def("exl3_gemv", &exl3_gemv, "exl3_gemv");
     m.def("exl3_gemm_num_kernel_shapes", &exl3_gemm_num_kernel_shapes, "exl3_gemm_num_kernel_shapes");
--- a/exllamav3/exllamav3_ext/quant/p2b_moe.cu
+++ b/exllamav3/exllamav3_ext/quant/p2b_moe.cu
@@ -1,0 +1,576 @@
+// Port of vcruz305/vllm-exl3 csrc/p2b_moe.cu (MIT, Copyright Mia AI Lab / vcruz305, see
+// THIRD_PARTY notices in that repository) to the exllamav3 tree, extended from one input row
+// per launch to a full row-expert slot table so a whole verify window (bsz * top_k slots)
+// runs as ONE cooperative launch instead of one per row.
+//
+// Four grid-synchronized phases over all active expert slots: input Hadamard, batched
+// gate/up GEMV (exl3_gemv_kernel.cuh MMA path), SwiGLU + down-input Hadamard, batched down
+// GEMV, then output Hadamard and weighted atomic accumulation. Grid size is occupancy-derived;
+// every phase grid-strides over its work items.
+//
+// Dimensions are runtime parameters (the reference build hardcoded hidden 4096 / inter 2048);
+// hidden and inter must be multiples of 128 (Hadamard warp tiling). MCG codebook only, K in
+// {2, 3, 4} and uniform across gate/up/down, matching MultiLinear pointer tables.
+
+#include <cuda_fp16.h>
+#include <torch/extension.h>
+#include <ATen/cuda/CUDAContext.h>
+#include <c10/cuda/CUDAGuard.h>
+#include <cooperative_groups.h>
+
+#include "../util.h"
+#include "../util.cuh"
+#include "exl3_gemv_kernel.cuh"
+#include "hadamard_inner.cuh"
+
+#include "p2b_moe.cuh"
+
+namespace cg = cooperative_groups;
+
+template <int bits, int cb, int CFG>
+__device__ __forceinline__ void run_gemv_tile(
+    const uint32_t* __restrict__ B32,
+    const half2* __restrict__ A2,
+    half* __restrict__ C,
+    int kslices,
+    int size_k,
+    int group,
+    int ntiles,
+    int warp,
+    int lane,
+    float (*sh_red)[1][32])
+{
+    // Verbatim extraction of exllamav3's exl3_gemv_kernel group-loop body (MMODE = 0,
+    // SMEM_STAGE = false), the production-validated m=1 decode GEMV. The reference port's
+    // tile was textually close but numerically wrong against this tree's kernel.
+    constexpr int WK = CFG == 0 ? 16 : 8;
+    constexpr int WNT = CFG == 0 ? 2 : 4;
+    constexpr int PF = CFG == 0 ? 4 : 2;
+    constexpr int FOLD = CFG == 0 ? 4 : 2;
+    constexpr int THREADS = WK * 32;
+    constexpr int COLS = WNT * 16;
+    constexpr int TWORDS = 8 * bits;
+    constexpr int LOADS = bits == 2 ? WNT / 2 : WNT;
+    constexpr int LSTRIDE = bits == 3 ? 24 : 32;
+
+    const int chunk = CEIL_DIVIDE(kslices, WK);
+    const int ks0 = warp * chunk;
+    const int myn = max(0, min(chunk, kslices - ks0));
+    const size_t slice_stride = (size_t) ntiles * TWORDS;
+
+    const bool r0_ok = lane < 4;
+    const half2 hzero = __half2half2(__ushort_as_half(0));
+
+    int x_src_a = 0, x_src_b = 0, x_s2 = 0;
+    if constexpr (bits == 2) {
+        int i1 = lane >> 1;
+        x_src_b = i1;
+        x_src_a = (i1 + 15) & 15;
+    } else if constexpr (bits == 3) {
+        int t_offset = lane << 3;
+        int b1 = (t_offset + 257) * 3;
+        int b2 = b1 + 21;
+        int i0 = (b1 - 16) / 32;
+        int i2 = (b2 - 1) / 32;
+        x_s2 = (i2 + 1) * 32 - b2;
+        x_src_a = i0 % 24;
+        x_src_b = i2 % 24;
+    }
+
+    const uint32_t* bp = B32 + (size_t) ks0 * slice_stride + group * WNT * TWORDS + lane;
+
+    auto ld_b = [&] (int i, int l) -> uint32_t {
+        if constexpr (bits == 3)
+            return lane < 24 ? __ldcs(bp + (size_t) i * slice_stride + l * LSTRIDE) : 0;
+        else
+            return __ldcs(bp + (size_t) i * slice_stride + l * LSTRIDE);
+    };
+
+    uint32_t pf[PF][LOADS];
+    #pragma unroll
+    for (int d = 0; d < PF; ++d)
+        if (d < myn)
+            #pragma unroll
+            for (int l = 0; l < LOADS; ++l)
+                pf[d][l] = ld_b(d, l);
+
+    FragC_h ch[WNT][2] = {};
+    float2 acc0[WNT][2] = {};
+
+    for (int ib = 0; ib < myn; ib += PF) {
+        #pragma unroll
+        for (int d = 0; d < PF; ++d) {
+            const int i = ib + d;
+            if (i >= myn) break;
+
+            uint32_t bw[LOADS];
+            #pragma unroll
+            for (int l = 0; l < LOADS; ++l)
+                bw[l] = pf[d][l];
+
+            if (i + PF < myn) {
+                #pragma unroll
+                for (int l = 0; l < LOADS; ++l)
+                    pf[d][l] = ld_b(i + PF, l);
+            }
+
+            const size_t a_col = (size_t) (ks0 + i) * 8 + (lane & 3);
+            FragB a01, a23;
+            a01[0] = r0_ok ? A2[a_col] : hzero;
+            a23[0] = r0_ok ? A2[a_col + 4] : hzero;
+            a01[1] = hzero;
+            a23[1] = hzero;
+
+            #pragma unroll
+            for (int t = 0; t < WNT; ++t) {
+                FragB f0, f1;
+                if constexpr (bits == 4) {
+                    uint32_t aw = __shfl_sync(0xffffffffu, bw[t], (lane + 31) & 31);
+                    exl3_gemv_ns::dq8_regs_4bits<cb>(aw, bw[t], f0, f1);
+                } else if constexpr (bits == 2) {
+                    const uint32_t w = bw[t >> 1];
+                    const int base = (t & 1) << 4;
+                    uint32_t bwv = __shfl_sync(0xffffffffu, w, base + x_src_b);
+                    uint32_t awv = __shfl_sync(0xffffffffu, w, base + x_src_a);
+                    exl3_gemv_ns::dq8_regs_2bits<cb>(awv, bwv, lane << 3, f0, f1);
+                } else {
+                    uint32_t awv = __shfl_sync(0xffffffffu, bw[t], x_src_a);
+                    uint32_t bwv = __shfl_sync(0xffffffffu, bw[t], x_src_b);
+                    exl3_gemv_ns::dq8_regs_3bits<cb>(awv, bwv, x_s2, f0, f1);
+                }
+
+                exl3_gemv_ns::mma_ab_h(a01, a23, f0, ch[t][0]);
+                exl3_gemv_ns::mma_ab_h(a01, a23, f1, ch[t][1]);
+            }
+
+            if ((d + 1) % FOLD == 0 || i + 1 == myn) {
+                #pragma unroll
+                for (int t = 0; t < WNT; ++t)
+                    #pragma unroll
+                    for (int f = 0; f < 2; ++f) {
+                        acc0[t][f].x += __low2float(ch[t][f][0]);
+                        acc0[t][f].y += __high2float(ch[t][f][0]);
+                        ch[t][f][0] = hzero;
+                    }
+            }
+        }
+    }
+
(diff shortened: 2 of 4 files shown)


creslinux:
# Validation of #356 (vcruz305 p2b port) — 4x3090

Build: `upstream/pr-356` (`b273034` = kernel `69c5a46` + integration `b273034`) on dev tip
`8473340`; CUDA 13.2.86 toolchain, torch 2.11.0+cu130, sm_86, `MAX_JOBS=16`, `--no-deps`, wall
425 s. Hardware: 4x RTX 3090 24 GB, layer split, qwen3.8-Flash-Next EXL3 4.05bpw. Side container;
serving restored after every window. Every number below is from the log named beside it.

Logs retained locally, available on request.

## Gate 1 — engagement, default off (`g1_on.log`, `g1_off.log`)

| arm | engagement print | 12-token run |
|---|--:|---|
| `EXL3_P2B_MOE=1` | exactly 1 | coherent, no errors |
| unset | 0 | runs, no errors |

## Gate 2 — MoE-level parity (`g2_parity_traced.log`)

Layers 0/24/47; real decode inputs at bsz 1/2/4/8; forward with the path off (dense) twice and on
(p2b) twice on the same tensors; a launch counter on `_p2b_forward_bszn_dev` labels which kernel
ran per cell. Relative error vs the dense output:

| layer | bsz1 | bsz2 | bsz4 | bsz8 | on-arm |
|---|--:|--:|--:|--:|---|
| 47 | 6.2e-3 | 1.20e-2 | 1.38e-2 | 1.35e-2 | p2b all rows |
| 24 | dropped | 7.4e-3 | 7.3e-3 | 1.54e-2 | p2b at bsz≥2 |
| 0  | dropped | 1.04e-2 | 1.28e-2 | 1.92e-2 | p2b at bsz≥2 |

Worst p2b-vs-dense (p2b-engaged) **1.92e-2**. Dense and p2b run-to-run spreads are 0.0 in every
reported cell. The two dropped bsz1 cells (raw 6.6e-7 / 2.3e-7) are not parity cells: the counter
shows the on-arm took dense, so both arms ran the same kernel; `g2_bsz1check.log` shows p2b does
engage at rows 1/2/3 when the method is called directly, so the non-engagement is specific to
re-entering `forward` at rows=1 in the harness. Layer 47 bsz1 did engage (6.2e-3). Magnitude is
deterministic 6e-3…1.9e-2: p2b is a per-slot fp32 GEMV with fp32 weighted atomic accumulation vs
the dense tiled mgemm with fp16 accumulate — the fp16 accumulation-order class; `p2b_stage_debug`
not triggered. A direct `_p2b_forward_bszn_dev` vs `bc.run_bszN out_d` parity was tried and
discarded (0.26–0.85) because the dense shared/tail is added outside `out_d`; the forward-level
comparison above includes shared on both arms.

## Gate 3 — unload / reload on a different split (`g3_unload.log`)

| step | cuda0 | cuda1 | cuda2 | cuda3 |
|---|--:|--:|--:|--:|
| load [16,22,22,22], gen (p2b on) | 15135 | 21620 | 21620 | 7309 |
| after `model.unload()` | 12 | 12 | 12 | 12 |
| reload [22,22,22,16], gen (p2b on) | 21670 | 21620 | 21959 | 8 |

Unload drops every device to ~12 MiB (p2b statics released). The reload re-shards onto **three**
cards (cuda3 empty, 8 MiB) — a different device set than first engagement — and generates cleanly
with p2b engaged. No illegal access.

## Gate 4 — decode: static window (rows/pass) and direct microbench

**Reconciling with the PR description.** The "+18-23%" / "+25%" quoted there were dynamic-window
throughput measurements on our September tree. On current dev, with the verify window held fixed,
the measured kernel effect is **−5.4% / −14.3% / −7.8% pass time at 4 / 5 / 6 rows** (about
+6% / +17% / +8% throughput). The description should therefore be read as "roughly 5-15% depending
on rows per verify pass", and vcruz305 is welcome to replace the quoted figures with these.

**Form 1, static draft window** (`dynamic_draft_tokens=False`, temperature 0, 3 runs/arm, bsz1).
Draft 3/4/5 give 4/5/6 rows per pass, so ms/pass is the kernel number and the three points are
directly comparable. Logs: `g4_static_off.log`/`g4_static_on.log` (draft 3),
`g4_static_off_d4.log`/`g4_static_on_d4.log` (draft 4),
`g4_static_off_d5.log`/`g4_static_on_d5.log` (draft 5).

| rows/pass (draft) | off ms/pass per run (mean) | on ms/pass per run (mean) | gain |
|---|---|---|---|
| 4 (3) | 36.02 / 36.71 / 36.63 (**36.45**) | 34.56 / 34.10 / 34.79 (**34.48**) | **−5.4%** |
| 5 (4) | 45.43 / 45.69 / 45.61 (**45.58**) | 39.07 / 38.99 / 39.15 (**39.07**) | **−14.3%** |
| 6 (5) | 48.39 / 48.44 / 49.00 (**48.61**) | 44.84 / 44.67 / 44.96 (**44.82**) | **−7.8%** |

Engagement checked before reading each on-arm row: print once; launch counts **17088** (4 rows),
**18384** (5 rows), **15360** (6 rows); off-arm **0** in every window. bsz8 (window 8x(1+draft) ≥ 32
rows > `MAX_BSZN`) is not on the p2b path in any window and is omitted.

**Gain as a function of rows/pass.** Off→on is 36.45→34.48 (−5.4%) at 4 rows, 45.58→39.07 (−14.3%)
at 5, 48.61→44.82 (−7.8%) at 6. The 5-row window is the largest because it straddles the dense
path's 4→5-row step below: the dense path pays that step once (4→5) and not again (5→6), while p2b
pays a smaller step, so the relative gap is widest at 5 rows and narrows by 6. This is the
pass-level reading of the microbench's per-call slope (dense ~+78 us/verified row over 1→5,
p2b ~+46 us), and it scales to the generator: with a dynamic window the accepted row count moves,
which is why the earlier dynamic-window ms/pass tracked the window rather than the kernel.

**Observation: the dense path has a 4→5-row cost step** (`g4_moebench.log`, per MoE call, us):

| rows | layer 47 | layer 24 | layer 0 |
|--:|---|---|---|
| 4 | 302.0 | 307.7 | 301.5 |
| 5 | 428.2 | 435.4 | 425.7 |
| step | +126.2 (+41.8%) | +127.7 (+41.5%) | +124.2 (+41.2%) |

p2b over the same step: 256.7→291.7 (+35.0, +13.6%), 258.3→296.1 (+37.8, +14.6%),
246.9→283.3 (+36.4, +14.7%). Stated as measured; this is a property of the current dense path,
not of this PR.

**Form 2, direct MoE microbench** (`g4_moebench.log`): `_p2b_forward_bszn` vs the dense
`bc.run_bszN` on captured decode inputs, CUDA events, synced, 50 iters, min of 3, three layers.

| rows | layer 47 dense→p2b (us) | layer 24 | layer 0 |
|--:|---|---|---|
| 1 | 116.2 → 108.6 (−6.5%) | 116.4 → 109.4 (−6.0%) | 116.3 → 108.8 (−6.5%) |
| 2 | 174.8 → 158.1 (−9.5%) | 176.4 → 159.2 (−9.8%) | 174.1 → 157.8 (−9.4%) |
| 4 | 302.0 → 256.7 (−15.0%) | 307.7 → 258.3 (−16.0%) | 301.5 → 246.9 (−18.1%) |
| 5 | 428.2 → 291.7 (−31.9%) | 435.4 → 296.1 (−32.0%) | 425.7 → 283.3 (−33.5%) |
| 8 | 572.4 → 455.9 (−20.3%) | 580.4 → 457.7 (−21.1%) | 553.0 → 437.3 (−20.9%) |

p2b is faster per call at every row count, −6% (rows 1) to −33% (rows 5). Dense is ~116 us at
rows 1 and adds ~78 us/verified row over 1→5; p2b is 108.6 us and adds ~46 us/row — the win is
mostly lower per-row cost with near-equal fixed cost. Generator-independent.

## Gate 5 — prefill 16k, regression check (`g5_off_traced.log`, `g5_on_traced.log`)

| arm | tok/s per run | mean |
|---|---|--:|
| off | 1615, 1612, 1615 | 1614 |
| on  | 1614, 1609, 1613 | 1612 |

p2b is decode-only; identical inside run spread, no regression.

## Gate 6 — p2b buffer VRAM at `MAX_BSZN` sizing (`g4_on_traced.log`)

81.6 MiB across 48 engaged layers = **1.70 MiB/layer** (`MAX_BSZN*topk = 80` slots: gate/up
intermediates, down and hadamard scratch, fp32 out/acc rows). `memory_allocated` delta after first
engagement: +73/+73/+74/+61 MiB on cuda0..3 (includes draft/cache growth).

## Device wrapper (`nowrap_repro.log`)

Reproduced with the wrapper removed (no #337), `torch.cuda.can_device_access_peer` true for
**all** pairs on this rig: the p2b run aborts `GPU assert: an illegal memory access` (exit 188,
surfaced at `graph.cu:40`). The `with torch.cuda.device(self.device)` scope keeps the cooperative
launch on the layer's own device; the wrapper stays correct with or without #337.

turboderp:
I'm unable to reproduce any speedup here for  Qwen3.8-Flash-Next, but Qwen3-30B-A3B shows some improvement on the 3090 and a little less on the 4090. This is probably worth replacing the bszN kernels with entirely, but a few issues:

- Expert biases are dropped (silently)
- SiLU only and no activation limits
- No staging of padded hidden dims
- Twelve static tensors per layer is a bit wasteful. `g_tensor_cache` exists to share statics between modules
- It loses some precision by downcasting to fp16 before activations and before the output hadamard
- GEMV autotuning doesn't reach it (maybe doesn't need to, not sure)
- The GEMV inner body is just the existing function but with a narrow configuration. I don't see a need to duplicate that code instead of just creating a templated instance
- I've switched to deterministic accumulation in `dev`, so this would be a slight regression in that regard. But easily addressable on its own.

creslinux:
- Expert biases: a real bug, and mine. _p2b_ok gates on quantised / gated / silu / act_limit / K / dims / offload but never checks for bias, so the path engages and silently ignores them. One clause, in the same place as the existing checks.
- Padded hidden dims and the twelve statics: also mine. The statics should go through g_tensor_cache; I sized them per module without looking for the shared path.
- Duplicated GEMV body: standalone was the only option in a vLLM plugin, but in-tree it should be a templated instance of the existing function — which also fixes autotuning not reaching it.

The fp16 intermediates and SiLU-only are the original kernel's design, so @vcruz305 call. The gate does at least refuse non-SiLU and activation limits rather than running them wrong.

On Flash-Next, was a draft model attached? The gain scales with rows per verify pass, so at bsz 1 with no speculation there is a single row and very little to see. On 8473340 I measured −6.5% per MoE call at 1 row against −31.9% at 5. Those numbers are pre-today's dev, so I am rebasing and re-measuring; I will post the current ones rather than argue from stale ones.

Happy to do the integration fixes if that is useful.

creslinux:
Re-measured on current dev, as promised.

Rebased `upstream/pr-356` onto `f64d5b2` (`ed34f82` + `5b1b06b`; the rebase branch `perf/pr356-v2` is local to the rig, not pushed). Two conflicts, both mechanical: `exllamav3_ext/bindings.cpp` (new dev's `had_r_128_batch` binding lands next to the p2b bindings — kept both) and `block_sparse_mlp.py::unload` (new dev's `batch_recon = None` dropped by the p2b hunk — kept it beside the p2b statics reset). No kernel-logic conflict. 4× RTX 3090, layer split, Qwen3.8-Flash-Next EXL3 4.05bpw, CUDA 13.2.86 / torch 2.11.

Engagement was confirmed before any on-arm number was read: with `EXL3_P2B_MOE=1` the path prints once and the launch counter is non-zero; unset, it is zero in every window, for both the p2b build with the env off and a separately built stock-dev control.

**Direct MoE microbench** — `_p2b_forward_bszn` vs dense `bc.run_bszN` on captured decode inputs, CUDA events, synced, 50 iterations, min of 3, µs per call:

| rows | layer 47 dense→p2b | layer 24 | layer 0 |
|--:|---|---|---|
| 1 | 116.5 → 109.6 (−5.9%) | 116.8 → 108.9 (−6.8%) | 116.4 → 110.0 (−5.5%) |
| 2 | 174.6 → 159.3 (−8.8%) | 174.9 → 158.2 (−9.5%) | 174.8 → 159.1 (−9.0%) |
| 4 | 302.5 → 252.6 (−16.5%) | 302.7 → 253.0 (−16.4%) | 290.8 → 245.0 (−15.7%) |
| 5 | 422.2 → 292.0 (−30.8%) | 424.7 → 292.4 (−31.2%) | 411.7 → 283.8 (−31.1%) |
| 8 | 564.1 → 443.7 (−21.4%) | 571.6 → 446.4 (−21.9%) | 554.9 → 437.7 (−21.1%) |

Old dev (`8473340`) for comparison, layer 47: 116.2→108.6, 174.8→158.1, 302.0→256.7, 428.2→291.7, 572.4→455.9. The p2b advantage is reproduced within ~1–2 points at every row (−16.5% vs −15.0% at 4, −30.8% vs −31.9% at 5, −21.4% vs −20.3% at 8), so folded Hadamard and the deterministic reduction did not absorb the win.

**Static draft window** — `dynamic_draft_tokens=False`, temperature 0, bsz 1. Six runs per arm across two passes with the arm order rotated between them, so ordering drift cancels. `off` is the paired control (p2b build, `EXL3_P2B_MOE` unset); the parenthetical is a separately built stock-dev control, which agrees within run-to-run spread:

| rows/pass (draft) | off, ms/pass (stock-dev ctl) | p2b on, ms/pass | gain |
|---|---|---|---|
| 4 (3) | 37.49 (37.45) | 35.35 | −5.7% |
| 5 (4) | 47.16 (47.77) | 41.03 | −13.0% |
| 6 (5) | 51.31 (51.04) | 46.66 | −9.1% |

The two forms reconcile: per-call savings × 48 layers is 6.25 ms/pass at 5 rows against 6.13 observed, and at 4 rows 2.20–2.40 predicted (layer 47 at the top of that range) against 2.14 observed.

The dense path's 4→5 row step survives. It is +119.7 / +122.0 / +120.9 µs per call on layers 47 / 24 / 0 (+39.6 / +40.3 / +41.6%), against +126.2 / +127.7 / +124.2 µs (+41.2–41.8%) on old dev. p2b's own step is +39 µs (+15.7%), which is why the relative gap peaks at a 5-row window and narrows either side of it.

That is also, I suspect, why a Flash-Next run can show nothing: without a draft model attached the verify window is a single row, where the saving is under 6% of the MoE call and invisible end to end.

The integration fixes from your review are still open in my tree. Say the word on the bias gate and I will put it on the fork branch for you to cherry-pick.

turboderp:
I had to reimplement this from scratch, since it needed a different approach for compatibility. Among other things, supporting different gate+up/down bitrates (without 3*7*7 kernel instances for that alone.) Also, wildly different performance profiles across GPUs and expert shapes and so on. Still improvements across the board, almost.

| Level | Model / shape | GPU | bsz 1 | bsz 4 | bsz 8 |
|---|---|---|---|---|---|
| Kernel | Qwen3-30B (2048×768, 128E, top-8, K4) | 5090 | 37.8 → 20.2 (−47%) | 85.7 → 55.1 (−36%) | 170.5 → 113.1 (−34%) |
| Kernel | Qwen3-30B -- | 4090 | 36.9 → 24 (−35%) | 84.6 → 80.4 (−5%) | 185 → 173 (−7%) |
| Kernel | Qwen3-30B -- | 3090 | 64.2 → 53.7 (−16%) | 181 → 168 (−7%) | 318 → 242 (−24%) |
| Kernel | gpt-oss (2944×2944, 32E, top-4, K4) | 5090 | 58 → 38 (−34%) | | |
| Kernel | gpt-oss -- | 3090 | 135 → 113 (−16%) | | |
| Kernel | 4096×2048, 128E, top-8, K4 | 3090 | 174.5 → 181.6 (+4%) | | |
| Decode | Qwen3-30B 4bpw | 5090 | 178 → 200 (+12%) | 426 → 492 (+15%) | 634 → 887 (+40%) |
| Decode | gpt-oss-20b 4bpw| 4090 | +7.6% | +13% | |
| Decode | Qwen3.5-122B 4bpw| PRO 6000 | +6% | +6% | +20% |
| Decode | Nemotron-3 Super 3bpw | PRO 6000 | +6.6% | +13% | +17% |
| Decode | GLM-5.3-Flash 2.05bpw | PRO 6000 | +6% | +8% | |
| Decode | LFM2.5-8B 4bpw | 5090 | +7.5% | | |
| Decode | Qwen3-30B, TP-2 | 4090+3090 | +14% | +51% | |
| Decode | GLM-5.3 8-layer stub 4bpw (4096×2048, 64E) | 4090 | +1.8% | −1.6% | +13% |
| Decode | GLM-5.3 8-layer stub 4bpw | 3090 | −2.2% | +17% | +40% |

Kernel results in us, decode is (relative) t/s.

PR #357, opened 2026-09-11
MoE: M=16/32/64 prefill row-tile dispatch

creslinux:
# Fused MoE prefill: M=16/32/64 row-tile dispatch

## Problem
The fused MoE re-runs the whole packed-weight load+decode pipeline once per 16-row M tile.
Nsight Compute on one full-routing layer call: 33% occupancy (1 block/SM, register- and
smem-limited), 4 warps/scheduler, 57% of cycles with no eligible warp, 25% SM / 13% DRAM, and a
22:1 ALU+FMA:tensor instruction ratio. It is a decode instruction stream stalled on barriers
(23%) and fixed-latency dependencies (19%), not MMA or bandwidth.

## Mechanism
A separate `exl3_gemm_kernel_inner_mt` holds `TILEBLOCKS_M` m16 row fragments and decodes each B
fragment once for all of them; the shared dense inner is byte-identical. Isolated marginal cost
per 16 rows: M=16 64.4 us, M=32 48.3 us, M=64 35.0 us. Dispatch by `token_count` (<=16 -> M=16,
<=48 -> M=32, else M=64 for N=128 / M=32 for N=256); the M=16 tier is inlined, M=32/64 run an
outlined whole-expert loop.

**Why a copy, not a parameter.** The `_mt` inner is a separate 975-line function rather than a
parameter on the shared inner because parameterising it reordered eight instructions in the dense
SASS; the copy is what keeps every existing model byte-identical. If you would rather have a single
parameterised inner and accept a scheduling perturbation on the dense path, that version is one
commit away.

## Results (4x3090, layer-split qwen3.8-Flash-Next EXL3 4.05bpw; all figures throughput)

| gate | dev tip | PR | |
|---|--:|--:|--:|
| full routing | 12.16 ms / 16.6 TF | 9.69 ms / 20.8 TF | **1.25x** |
| prefill 16k | 1498 tok/s | 1702 tok/s | **1.14x** |
| prefill 64k | 1796 tok/s | 1919 tok/s | **1.07x** |
| decode bsz1 / bsz8 | ~58 / ~152 tok/s | unchanged | 1.00x |

Only `k4_n128_cb2` (N=128, K=4, mul1) is exercised by this checkpoint, all 48 layers, so that
tier is measured on real weights. The N=256 tier is measured on seeded synthetic trellis.

## Correctness
- per-expert parity: bit-identical vs the current kernel; full routing max abs 1.9e-9.
- dense path: all 36 dense objects (12 top-level + 24 comp-units) have identical SASS.
- N=256 synthetic tier: bit-identical base vs PR at every C.
- 16k logits: top-3 token ids identical across base and PR; values are within the model's own
  run-to-run nondeterminism (base-vs-base max abs 3.29 > base-vs-PR 2.72; n=10 per arm), so the
  id agreement is the end-to-end gate, not a bitwise logits comparison.

## Rejected
- Occupancy (2 blocks/SM, 64 regs): residency rose to 57% but full routing regressed to 47 ms.
- N-split (own N range, no K partial): 12.71 ms at `TILESIZE_N=128` (only 5 N tiles for the
  640-wide gate over 8 blocks).
- Uniform M=64: 12.87 ms (median expert holds 13 rows; 3 of 4 fragments wasted).

Build cost: +15 s wall, +14 MB `.so` (same `MAX_JOBS=16`).

### DIFF
--- a/exllamav3/exllamav3_ext/quant/exl3_gemm_inner.cuh
+++ b/exllamav3/exllamav3_ext/quant/exl3_gemm_inner.cuh
@@ -736,3 +736,978 @@ void exl3_gemm_kernel_inner
         }
     }
 }
+
+
+// Multi-row-M prefill variant used only by the fused MoE kernel. Kept separate from the
+// function above so the dense/decode instantiations (TILESIZE_M == 16) are untouched and
+// remain byte-identical. TILEBLOCKS_M row fragments share each dequantized B fragment.
+template<EXL3_GEMM_T_ARGS, bool shmem_out_had>
+inline __device__
+void exl3_gemm_kernel_inner_mt
+(
+    const half* __restrict__  A,
+    const uint16_t* __restrict__ B,
+    void* __restrict__ C,
+    const int size_m,
+    const int size_k,
+    const int size_n,
+    int* __restrict__ locks,
+    const half* post_scale,
+    int size_n_stride = 0     // full width of B and C when computing a column slice (0: = size_n)
+)
+{
+    // Row fragments: one m16 fragment per 16 rows. TILEBLOCKS_M == 1 is the original dense/decode
+    // shape (byte-identical codegen); > 1 is the prefill MoE shape, where several row fragments
+    // share each dequantized B fragment instead of re-running the whole B pipeline per 16 rows.
+    const int TILEBLOCKS_M = TILESIZE_M / 16;
+    if (size_n_stride == 0) size_n_stride = size_n;
+    // Column blocks of the full-width B row: slices index B relative to their own column offset,
+    // but a k-tile row still spans the whole matrix
+    const int blocks_n_full = size_n_stride / 16;
+    const int TILEBLOCKS_K = TILESIZE_K / 16;
+    const int TILEBLOCKS_N = TILESIZE_N / 16;
+    // const int FRAGS_M = TILEBLOCKS_M;
+    const int FRAGS_N_PER_WARP = 2 * TILEBLOCKS_N / (EXL3_GEMM_BASE_THREADS / 32);
+
+    const int sh_a_stage_size = TILESIZE_M * TILESIZE_K;                         // in halfs
+    const int sh_b_stage_size = TILEBLOCKS_K * TILEBLOCKS_N * 256 / 16 * bits;   // in uint16s
+    const int sh_c_size = MAX  // in floats
+    (
+        4 * EXL3_GEMM_BASE_THREADS * FRAGS_N_PER_WARP * TILEBLOCKS_M,
+        shmem_out_had ? TILESIZE_N * TILESIZE_M : 0
+    );
+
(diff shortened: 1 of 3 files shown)


creslinux:
I need to rebase this MR as dev is no longer at 8473340. 
Of the eleven commits landed today, and four of them are in the MoE path this MR modified - including:

- a faster mixed-mode fp16 plus fp32 GEMM for the reconstruct path, explicitly aimed at consumer GPUs, which is the exact sibling path this M-tile work was measured against
- deterministic reduction switched on by default, with folded Hadamard
- a batched reconstruct path and a worst-case measurement during autosplit load

I hope to have time later today to rebase

turboderp:
I don't think there's any need. I'm reimplementing it on my end with three-tiered launch splitting between M=16, 32 and 64, split compilation units for K=1..6 and integration into CPU streaming mode. Just need to figure out how to squash everything together with this PR. Github is hard. :|

turboderp:
Isolated kernel latencies, Qwen3.8 shape, tiered versus single launch

.         | M=32  | M=48  | M=64  | M=128 | mixed M=1..128 
----------|-------|-------|-------|-------|----------------
4090      | 0.71x | 0.88x | 0.73x | 0.72x | 0.85x
3090      | 0.73x | 0.90x | 0.73x | 0.71x | 0.91x
5090      | 0.67x | 0.73x | 0.60x | 0.61x | 0.80x
Pro 6000  | 0.68x | 0.64x | 0.53x | 0.53x | 0.77x

End-to-end 4k prefill t/s:

.                                     | Before   | After    | Speedup
--------------------------------------|----------|----------|----------
Pro 6000, Qwen3.8-NF 3.05 bpw         | 6864 t/s | 7337 t/s | +6.8%
Pro 6000, GLM5.3-F 2.05 bpw           | 3492 t/s | 3917 t/s | +12.2%
2x3090+avx2(8t) Qwen3.8-NF 3.05 bpw   | 1733 t/s | 1837 t/s | +6.0%

creslinux:
M=128 winning everywhere is the part I never reached. Four fragments spilled at M=64 (STACK 792) and I only got it down to 248 by single-buffering A, since the load_frags→matmul order means A never needs the full pipeline depth. I assume the split compilation units for K=1..6 are what frees the registers, the runtime switch over K being most of the live set.

On squashing: maintainer edits are enabled on this PR, so you can push straight onto creslinux:perf/moe-prefill-pr and it will update in place with the history underneath. Or land yours and close this one — either is fine, no need to contort git on my account.

Happy to run your version on 4x3090 layer split if a second data point helps; your 3090 row is two cards with CPU offload, mine is four cards without.

turboderp:
Okay, I think I figured out how to github now. I just merged it into the void somehow. o_O

creslinux:
Appreciate the extra merging effort for attribution — didn't expect that, thank you. Nice result on the four-card spread.

PR #362, opened 2026-09-12
Fix/dsa optional triton import 309

### LINKED ISSUE #309: bc_dsa.py imports triton-only kernels unconditionally, so exl3 1.4.x cannot import on a triton-less machine
## Summary

`exllamav3/modules/attention_fn/bc_dsa.py` imports names from `dsa_triton` at module
scope, unconditionally. `dsa_triton.py` itself guards the triton import correctly and
defines those names only inside `if has_triton:`. So on a machine without triton,
`import exllamav3` fails — and because `architecture/architectures.py` imports every
architecture at module level, the failure is not confined to DeepSeek: **no model of any
architecture can be loaded.**

This matters on Windows in particular, where official `triton` has no distribution at all.

## Reproduction

Windows 11, Python 3.10, `torch==2.10.0+cu128`, `exllamav3==1.4.4+cu128.torch2.10.0`,
no triton installed:

```python
>>> from exllamav3 import Config
>>> Config.from_directory("path/to/any-exl3-model")
```

Import chain:

```
exllamav3/architecture/architectures.py   (imports every architecture at module level)
  -> deepseek_v4
    -> dsv4
      -> modules/attention_fn/bc_dsa.py:9
         from .dsa_triton import _dsa_attn_split_kernel, ...
```

`dsa_triton.py` does the right thing:

```python
try:
    import triton
    import triton.language as tl
    has_triton = True
except ImportError:
    has_triton = False

if has_triton:
    @triton.jit
    def _dsa_attn_split_kernel(...):
        ...
```

...but `bc_dsa.py` imports those names regardless, so the guard cannot take effect.

## Versions affected

Confirmed present in **1.4.0, 1.4.2, 1.4.3 and 1.4.4**.

## Suggested fix

Guard the import in `bc_dsa.py` the same way `dsa_triton.py` guards its own, and defer the
failure to the point of use — so a machine without triton can still load architectures that
do not need it. Something like:

```python
try:
    from .dsa_triton import _dsa_attn_split_kernel, ...
    _HAS_DSA_TRITON = True
except ImportError:
    _HAS_DSA_TRITON = False
```

...with a clear error raised only when a DSA path is actually taken.

## Workaround

Installing `triton-windows` supplies a `triton` module and the import succeeds; with that
in place 1.4.4 loads and generates normally here (23 EXL3 model directories, text and
vision, including the vision tower path).

⚠ One caveat worth recording for other Windows users: with `triton-windows` present,
additionally installing the declared dependency `flash-linear-attention` makes
`import exllamav3` fail again, this time inside `fla/ops/simple_gla/parallel.py` under
triton's JIT source parser (observed at triton-windows 3.5.1 and 3.7.1). Omitting
`flash-linear-attention` works because every `fla` import inside exllamav3 is guarded;
a GatedDeltaNet model (Qwen3.5-9B) loads and generates without it via `gdn.cu`.

Komeiji-Shiki:
## Summary

Fixes #309.

`dsa_triton.py` defines its JIT kernel symbols only when `has_triton` is true, but `bc_dsa.py` imports three of them unconditionally. This raises `ImportError` while the architecture registry is being imported, including for architectures that do not use DeepSeek DSA.

This change:
- imports the three DSA kernels only when `dsa_triton.has_triton` is true;
- disables `bc_dsa_enable` when Triton is unavailable;
- makes both single-job and batched BC builders decline immediately without Triton, including with `EXL3_BC_DSA_DEBUG=1`;
- adds CPU-only import/availability regression tests covering the missing and available dependency cases, the environment switch, both builders, and propagation of genuinely missing kernel symbols when Triton is present.

The existing Windows dependency declaration and GPU kernel implementations are unchanged. The Python/FLA compatibility caveat mentioned in the issue is outside this patch's scope.

## Verification

Based on upstream `dev` at `f64d5b2a43e1a1094930a18c0de86b5476335562`.

The regression tests reproduce the original `ImportError` on the unmodified source. With this patch:

```text
python -m unittest discover -s tests -p test_bc_dsa_import.py -v
Ran 5 tests
OK

PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_bc_dsa_import.py
5 passed, 15 subtests passed

python -m py_compile exllamav3/modules/attention_fn/bc_dsa.py tests/test_bc_dsa_import.py
git diff --check
```

Tests were executed on Linux with Python 3.13.5. They load the complete, real `bc_dsa.py` module using isolated dependency stubs; they exercise import and dispatch-availability behavior without compiling CUDA or running model inference. Full-package loading, Windows execution, and GPU inference have **not** been tested in this environment.

### DIFF
--- a/exllamav3/modules/attention_fn/bc_dsa.py
+++ b/exllamav3/modules/attention_fn/bc_dsa.py
@@ -305,8 +305,8 @@ class BCDsa:
 
     def run(self, x, rs, rsl, bt_row = None):
         """x (1, seq, hidden) fp16 contiguous, bt_row (1, npr) i32 device block-table row of
-        this job (paged pools). Returns y (1, seq, hidden) fp32 (a static: consume before the
-        next BC call), or None to decline (host-side ring maintenance needed this
+        this job (paged pools). Returns y (1, seq, hidden) fp32 (a static: consume before
+        the next BC call), or None to decline (host-side ring maintenance needed this
         step)."""
         seq = x.shape[1]
         pos = rs.position
@@ -414,7 +414,7 @@ class BCDsaBatch:
                 m.indexer._build_fused()
             fan_lins += [m.indexer.wkv, m.indexer.wgate]
         fan_inner = [l.inner for l in fan_lins]
-        self.fan_ns = [l.out_features for l in fan_inner] if False else [l.out_features for l in fan_lins]
+        self.fan_ns = [l.out_features for l in fan_lins]
         q_lora = m.q_a.out_features
         self.q_lora = q_lora
         if not (
--- a/exllamav3/modules/attention_fn/bc_dsa.py
+++ b/exllamav3/modules/attention_fn/bc_dsa.py
@@ -6,7 +6,11 @@ from ...ext import exllamav3_ext as ext
 from ...constants import PAGE_SIZE
 from ...util.tensor import g_tensor_cache
 from .bc_attn import _compile_kernel
-from .dsa_triton import _dsa_attn_split_kernel, _dsa_attn_combine_kernel, _dsa_indexer_fewq_kernel
+from .dsa_triton import has_triton
+
+# These kernel names are only defined when Triton is available.
+if has_triton:
+    from .dsa_triton import _dsa_attn_split_kernel, _dsa_attn_combine_kernel, _dsa_indexer_fewq_kernel
 
 """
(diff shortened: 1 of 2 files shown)


turboderp:
This seems to already have been resolved differently, since I think v1.4.4 which this is based on. Triton is now a hard dependency, and pyproject.toml pulls `triton-windows` on Windows. The try-imports are vestigial from when the library primarily used flash-attention-2, but that's gone now, so I'm cleaning up those remaining sites instead.

PR #364, opened 2026-09-12
sc_measure: add layer streaming

UnstableLlama:
`sc_measure.py` currently requires the full model to fit on one GPU. This adds layer streaming so larger models can be measured by loading one top-level module at a time and keeping cached states in system RAM.

Following the approach in `conversion/measure_model.py`, perturbed states advance through subsequent modules together, sharing each module load. Pending states are grouped into passes using a RAM budget. `--max-sys` overrides the default of half the available RAM after the reference pass.

The default loading mode is `auto`: estimate resident VRAM requirements and choose full-model loading when it fits, otherwise stream. An OOM during the initial full-model load also falls back to streaming. `--streaming` and `--no-streaming` force either mode. The noise generation logic and resident suffix-forward logic are preserved.

**Validation:**

- Full Qwen3.5-0.8B shaped-noise measurement against the original upstream script: all 302 KL scores across 151 tensors, realized noise values, and weight norms matched exactly. The maximum absolute difference in `inj_rfn` was approximately 1.87e-9. Upstream resident loading took 11:55 with 4.69 GiB peak VRAM; forced streaming took 15:21 with 3.01 GiB peak VRAM.

- A 49-tensor slice of Qwopus3.8-27B on one RTX 3090 compared fan-out against an earlier, unpublished simple streaming implementation, using the same calibration inputs and library build with a warm page cache. Time decreased from 46:01 to 26:21 (1.75x), with exact agreement in the recorded measurement values. Forcing eight passes with `--max-sys 3` took 27:20 and also matched exactly. These timings compare two streaming implementations, not streaming against upstream resident loading.

### DIFF
--- a/sc_measure.py
+++ b/sc_measure.py
@@ -5,11 +5,13 @@ import json
 import math
 import zlib
 from collections import defaultdict
+from contextlib import contextmanager
 import torch
 
 from exllamav3 import Config, Model, Tokenizer
 from exllamav3.util.file import disk_lru_cache
 from exllamav3.util.measures import compute_kl_div
+from exllamav3.util.tensor import g_tensor_cache
 from exllamav3.modules.linear import Linear
 from exllamav3.modules.quant.fp16 import LinearFP16
 from exllamav3.modules.quant.exl3_lib.quantize import finalize_capture_H, get_hadamard_dt, had_k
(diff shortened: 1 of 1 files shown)


UnstableLlama:
oops accidentally deleted the branch earlier and closed this x.x

PR #365, opened 2026-09-12
Make the extension build and load under ROCm/HIP

sdougbrown:
> [!note]
> this is going to result in a series of PRs that get a fully working and performant Qwen 3.8 Flash Next working on a 2x AMD R9700 machine at 3bpw (~65tok/s, 500k+ q4 kv pool 😎)
> 
> I'm sorry that it ends up being a lot of lines

Make the extension build and load under ROCm/HIP.

The C++ extension currently assumes CUDA in its build configuration and binding layer.
This PR is the foundation that lets it compile, link, and load with a HIP toolchain
without changing anything on CUDA.

Changes:

- `exllamav3/exllamav3_ext/build_config.py`: central source list used by both setup.py and
  ext.py, with an exclusion set for sources that have no HIP-portable form yet
- `setup.py` / `exllamav3/ext.py`: ROCm compiler flags (USE_ROCM, arch list, two-line
  `-Wno-register`) and source selection
- `exllamav3/exllamav3_ext/compat_rocm.cuh` and `compat.cuh`: HIP polyfills for the
  CUDA-specific constructs the extension relies on
- `exllamav3/exllamav3_ext/bindings.cpp`: USE_ROCM split in the binding registration
- `exllamav3/exllamav3_ext/rope.cu`: ROCm unweighted-scale RMS-norm path
- `exllamav3/exllamav3_ext/graph.cu`, `cuda_drv.*`: graph and driver shims for HIP

The compat layer is derived from open PR #283 by Calandracas606; Daniel Martinez's
commits are preserved with attribution and the branch rebases cleanly onto #283 if it
merges first.

Validation: on AMD Radeon AI PRO R9700 (gfx1201) with ROCm 7.2.4 and torch
2.12.0+git6bbd260 (HIP 7.2.53211):

    env -u CUDA_HOME PYTORCH_ROCM_ARCH=gfx1201 \
        TORCH_EXTENSIONS_DIR=$HOME/.cache/exllamav3-upstream-stage-ext \
        ROCR_VISIBLE_DEVICES=0 /home/douglasbrown/vllm-test-env/bin/python -c 'import exllamav3'

reports that the extension module loads. Later stacked PRs add the runtime fallbacks and
kernel portability that make it run a model.

PR #366, opened 2026-09-12
Add a PyTorch fallback layer for ROCm-excluded extension functions

sdougbrown:
Stacked on #365 (build/compat); this branch includes that PR's commits.

> [!note]
> i looked at a way to stack this in a way that makes it easier to review but it's kind of annoying from a fork. i could maybe set up a proper stack *on* my fork to make the diff easier to parse?

Add a pure-PyTorch fallback layer so the extension is fully runnable on ROCm when CUDA-only
functions are compiled out. With this PR (after #365) a model loads and generates correctly
on ROCm, using the reconstruct+hgemm path and torch-composed kernels at the documented
"correctness over speed" quality bar.

Changes:

- `exllamav3/ext_fallbacks.py` and `exllamav3/ext.py`: PyTorch implementations matching the
  C++ extension signatures, monkey-patched onto the module for every symbol absent from the
  build, plus a `_BCNone` sentinel so BC_* constructors degrade to None
- `exllamav3/modules/block_sparse_mlp_routing.py`: torch-composed fallbacks for every routing
  kernel absent from the build, matching each kernel's contract (top-k tie semantics, the
  activate_all bias fold, selection-only vs weight bias)
- availability gates in `dsv4.py`, `hyperconnections.py`, `ple.py`, `ngram_embedding.py`,
  `bighead_scalar.py`, and `util/memory.py` so fused paths route to eager implementations
- `exllamav3/generator`: n-gram drafting support validation; fused-sampler path now falls
  back to the step-by-step implementation when the fused kernel is absent
- tests that drive directly-at-CUDA-only functions now skip when those bindings are absent

Validation on gfx1201 (env as in #365): `pytest tests/` for the HIP-relevant files
(test_ext_norm_, test_hip_*, test_ext_routing_, test_moe_cpu_*, test_sampler, etc.) passes;
the stack's full suite is 914 passed / 29 skipped / 0 failed once #366 stacks on top.


sdougbrown:
`👾 AI Agent`

Readable stacked diff (this layer only, base `pr1`): fork mirror [PR2](https://github.com/sdougbrown/exllamav3/pull/2). Whole fork series: [PR1](https://github.com/sdougbrown/exllamav3/pull/1) / [PR2](https://github.com/sdougbrown/exllamav3/pull/2) / [PR3](https://github.com/sdougbrown/exllamav3/pull/3).

PR #367, opened 2026-09-12
Port recent kernel additions for HIP builds

sdougbrown:
Stacked on #366 (fallback layer); this branch includes that PR's commits.

Make the kernel sources the reconstruct and fused GEMM paths rely on build and run
correctly under hipcc. PR #365 and #366 make the extension load and fall back where a
kernel is absent; this PR fixes the native kernel code itself so the ROCm build is stable
and correct on the paths that do run.

Changes:

- `exllamav3/exllamav3_ext/quant/codebook.cuh`: replace the raw NVIDIA `lop3.b32` inline PTX
  with the equivalent `c ^ (a & b)` form. On AMDGPU, `"r"`-constrained operands allocate
  scalar registers for lane-varying dequant data, which either fails codegen or silently
  miscompiles it
- `exllamav3/exllamav3_ext/quant/reconstruct.cu` and `dsv4_pool_quant.cu`: launch batched,
  runtime-dispatched kernels through a local binding so hipify can rewrite them
- `exllamav3/exllamav3_ext/hgemm_f16acc.cu`: stub on ROCm (its mma.sync/cp.async path has no
  RDNA equivalent) so `hgemm_recon`/`hgemm_batched` degrade to cuBLAS instead of failing to link
- `exllamav3/exllamav3_ext/cache/lmq.cuh`: annotate the clamp helper for device compilation
- `exllamav3/exllamav3_ext/bindings.cpp`: register the hgemm family, batched hadamard,
  rewind jobs, ngram helpers, and dry penalty in the ROCm bindings branch where their TUs
  compile on HIP

Batched reconstruct kernels (`reconstruct_batch`/`reconstruct_had_batch`) stay CUDA-only
behind the existing binding gates pending resolution of a gfx1201 backend codegen issue;
their absence degrades to the per-expert reconstruct path, not a crash.

Validation on gfx1201 (env as in #365): the stack's full suite is 914 passed / 29 skipped /
0 failed.

One correctness follow-up, also extracted separately: the fused RoPE normalization reduction
race fix lives in its own PR (#363).


sdougbrown:
`👾 AI Agent`

Readable stacked diff (this layer only, base `pr2`): fork mirror [PR3](https://github.com/sdougbrown/exllamav3/pull/3). Whole fork series: [PR1](https://github.com/sdougbrown/exllamav3/pull/1) / [PR2](https://github.com/sdougbrown/exllamav3/pull/2) / [PR3](https://github.com/sdougbrown/exllamav3/pull/3).

PR #371, opened 2026-09-13
Fix CPU page cache cleanup on generator close

mdenomme24:
## Summary

`CPUPageCache` starts a background allocation thread whose bound method keeps the cache object alive indefinitely. The cache's `segments` then retain references to GPU KV cache tensors after the model/generator is closed.

This can leave GPU VRAM allocated after model unload when CPU KV cache paging is enabled, potentially causing subsequent model loads to run out of memory.

## Fix

* Make the CPU page-cache allocation worker stoppable.
* Add `CPUPageCache.close()` to stop the worker and release its tensor references.
* Call `CPUPageCache.close()` from `AsyncGenerator.close()`.
* Ensure the worker has exited before the cache's tensor references are cleared.

## Testing

Reproduced the VRAM leak on the unmodified upstream code and verified that the same reproduction passes with this change, with GPU memory returning to the expected baseline after generator cleanup.


turboderp:
Thanks. :+1: 

I changed it to use a weakref so the pages shouldn't leak even if the Generator isn't explicitly closed.

PR #376, opened 2026-09-15
MoE CPU split: can_defer_load reads infer_params.moe_cpu_split, not the env var (static placement from -mcs produced garbage)

midagedev:
## Summary

`BlockSparseMLP_CPU.can_defer_load()` decides whether the module's tensors may be filled by the loader's deferred pass. A static expert placement (`EXL3_MOE_CPU_SPLIT_STATS` with `EXL3_MOE_CPU_SWAP=0`) permutes the router-side tensors right after the module loads, so the guard returns `False` in that case — but it detected the split by reading the `EXL3_MOE_CPU_SPLIT` env var, which `-mcs` / `--moe_cpu_split` never sets (that flag sets `infer_params.moe_cpu_split`; the env is only that field's construction-time default, `model/config.py`). On the CLI path the guard was therefore dead: the deferred router fill landed in the pre-permutation tensor and was lost, and the model generated garbage while reporting a normal token rate.

This PR makes the guard read `infer_params.moe_cpu_split`, the source `cpu_maybe_split_load` already treats as authoritative, and adds a unit test for the guard. No behaviour change without a stats file.

## Reproducer

GLM-5.3-Flash, `turboderp/GLM-5.3-Flash-exl3` 4.05 bpw, exllamav3 1.5.0 wheel (`block_sparse_mlp_cpu.py` is byte-identical between the v1.5.0 tag, `master` and `dev` at 1996609), one 48 GB + one 24 GB card, 32-thread AVX2 host. Stats file = per-layer router selection counts from a 10-run dynamic-placement probe (the runtime's `_split_hist`). Greedy decode, 400 tokens, three prompts, `-mtp` with the draft depth at 1:

```
EXL3_MOE_CPU_SWAP=0 EXL3_MOE_CPU_SPLIT_STATS=stats.json \
  python bench.py -m GLM-5.3-Flash-exl3-4.05 -gs 44,21 -mcs 250 -mct 32 -cs 32768 -mtp
```

| arm | output | tok/s (3 prompts) | draft accepted / rejected |
|---|---|---|---|
| unpatched | garbage (`1. 1. 1. …`, prompt echoed back) | 21.4 / 21.2 / 20.7 | 199/2 (degenerate) |
| unpatched + `EXL3_MOE_CPU_SPLIT=250` set to match `-mcs` (env guard fires) | coherent | 22.7 / 18.4 / 19.5 | 73–88 % |
| this patch, no env var | coherent | 22.8 / 19.7 / 19.2 | 67–78 % |

Each arm registers all 43 MoE layers on the worker as `static … [38..288)`; the only difference is whether the router permutation sees the loaded tensor.

## Test

`tests/test_moe_cpu_split_defer_guard_.py` builds the module without weights (`__new__`, empty `modules`, a stub `config.infer_params`) and checks the three cases: split from `infer_params` + stats file → no defer; stats file without a split → defer; split without stats → defer. On the unpatched 1.5.0 wheel the first case fails, on the patched file all three pass (run on the machine above; the module imports the compiled extension so it needs an installed wheel).

## Relation to #315

#315 (`Precomputed MoE expert-placement profiles`) contains the same fix inside its `can_defer_load` hunk, widened for its own `EXL3_MOE_PROFILE` mode. That PR is large and its shape is still being decided; this one is the bug fix alone, for the `-mcs` + stats-file path that exists in the current release. If #315 lands first this change is subsumed by it and can be closed. Credit to @anoane for reaching the same diagnosis independently.

## Verification boundary

One machine, one model, the 1.5.0 wheel with the patched file dropped in; not run under tensor parallel, `--moe_cpu_offload`, or the conversion call sites of `can_defer_load` (they see the same `config.infer_params`, unchanged default 0).

AI assistance was used while preparing the patch and validation. I reviewed the change and am responsible for it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


turboderp:
:+1:

Issue #285, opened 2026-08-14
DFlash draft cache allocates full-size KV for all-SWA layers — blocks 131k speculative decoding on 12 GB; partial patch included


msad12:
**Setup:** RTX 3060 12 GB · exllamav3 1.4.2+cu128.torch2.9.0 · TabbyAPI · target `turboderp/Muse-Glimmer-30B-exl3` 3.00bpw · draft `turboderp/Muse-Glimmer-30B-assistant-exl3` 2.00bpw (611 MB) · `max_seq_len 131072` · cache Q4.

**Problem:** The DFlash draft model's layers are all `sliding_attention` (window 2048), but `architecture/dflash.py` instantiates plain `Attention` for every layer, so the draft's KV cache is allocated at full `cache_size`. At 131072 that is ~0.7 GiB for a 611 MB draft — together with workspace this makes target+draft OOM on 12 GB at long context, even though the pieces individually fit. (Same disease independently reported in other engines: Anbeeld/beellama.cpp#98.)

**What I tried:** mirroring the main model's pattern (`muse_glimmer.py`, `attn_cls = Attention if swa_full or not is_swa else SlidingAttention`) inside `dflash.py`:

```python
is_swa = config.layer_types[idx] == "sliding_attention"
attn = (SlidingAttention if is_swa else Attention)(
    ... # unchanged kwargs, sliding_window = config.sliding_window if is_swa else -1
)
```

**Result:** VRAM problem solved — target 2.5bpw + draft now loads at 131k in 10.4 GiB (was: hard OOM). But generation fails immediately:

```
File ".../backends/exllamav3/model.py", line 1402, in generate_gen
  async for result in job:
...
  instance].update_kv_direct(cache_seqlens, block_table, k, v, length)
KeyError: (0, 0)
```

The DFlash drafting path writes draft KV via `Cache.update_kv_direct(...)` addressed to standard paged layers; with `SlidingAttention` the layers register as recurrent/ring state (`sliding_attn.py` / `SWALayerState`), which doesn't expose that interface. Bridging the DFlash direct-KV update path onto the ring cache (as PR #253 did for Gemma4 SWA) is beyond what I can safely patch from outside.

[...]

PR #377, opened 2026-09-16
spark2_5: use SlidingAttention (windowed KV ring) for sliding-window layers

dfein38347g:
## Problem

`spark2_5` (added in #345) builds **all 36 layers with plain `Attention`**, so every layer gets the full-width paged KV cache. The 27 sliding-window layers (window 512) only *read* the last 512 rows — the rest is allocated but never used. On a 6 GB card that's ≈ **1.6 GiB of dead KV at 50k ctx** (measured on 0.9: 16k→32k ctx cost = +416.4 MiB).

The other six hybrid architectures in this release already use the windowed storage: **`SlidingAttention`** (`modules/sliding_attn.py`, present since v1.0.0) — a per-batch-slot fp16 ring of just `window + 2 overprovision pages` (≈ 0.59 GiB for 9 full layers + ~0.11 GiB of rings at 50k vs ≈ 2.3 GiB today), with checkpoint/stash/rollback support for speculative decoding. Callers: `gemma4.py`, `muse_glimmer.py`, `step3_5.py`, `laguna.py`, `afmoe.py`, `gpt_oss.py`. Spark is the only one paying full-width.

## Changes (no C++; pure Python)

1. **`architecture/spark2_5.py`** — per-layer class pick from `config.layer_types` (mirrors the gemma4/muse pattern; `layer_types` was already parsed and validated by the config). Identical per-layer arguments for both classes.
2. **`modules/sliding_attn.py`** — adds fused-QKV support (`key_fused_qkv` + `fkey`/`frange` slicing, exactly as `Attention` does) plus the `transpose_qkv`/`qbits_key` knobs, because Spark's EXL3 checkpoint carries one fused `q_k_v_proj` tensor. **No behavior change for the six existing callers**: `ftranspose_after_load` defaults to `True` and `qbits_key` to `bits` in `Linear`, and the separate `key_q/key_k/key_v` paths are byte-identical.

## Why this is safe

- The ring stores *exactly* what the attention kernel already reads (the last `sliding_window` rows), so the attention math is unchanged — the kernels were already windowed; only the storage was oversized.
- Ring size = `ceil((window + 2·PAGE_SIZE)/PAGE_SIZE)·PAGE_SIZE` fp16 rows per slot, attached as a recurrent layer (the existing `SWALayerState` machinery, same as gemma4 etc.).
- The DFlash drafter is untouched — its full-size cache issue is separate (#285, which cross-references the same disease in beellama.cpp#98).
- No config/CLI surface added; the existing `layer_types` json (or its absence → all-full, unchanged behavior) is the only input.

## Validation plan (run on a 6 GB card, 0.9 box)

1. Golden A/B, temp 0, ~10 prompts × 3 seeds: old vs new — expected byte-identical (attn math unchanged; only storage differs).
2. VRAM probe: weights + 9 full layers + rings ≈ 3.0 GiB → expect the ctx ceiling to rise from 53248 rows (~5.5 GiB) several-fold (predicted new OOM point ≈ 11.0–11.3 KiB/row → ~ 2× the rows per MiB, i.e. ~100–130k rows before the 5600 MiB cap).
3. 24 h soak at the new limit: quality, no ring-shift artifacts, spec-dec (DFlash) still clean.


dfein38347g:
FYI for the merge decision: this ring is the piece that lets us run Spark-X2.5-4B at 100k ctx (101k-token loads, ~5.0 GB peak) under a 5.6 GB cap on a 6 GB RTX 3050 — our production config on that box, running daily with no issues (see the close of #378 for the re-verification of the non-native-window question). Nothing further needed from us on our side.

turboderp:
:+1:

PR #380, opened 2026-09-17
DFlash2 on the existing DFlash path — integration, CPU-only CI, and a losslessness proof

rktchip:
Parallel integration of this + #334, built before your merge. Your tree supersedes it — not reopening. Four things from the measurement run, on RTX 5090, Qwen3-27B EXL3 3.5bpw + published DFlash2 5.0bpw, same seed, all paired same-prompt.

**1. `-dc` crop is negative EV — applies to the merged #334 path.** 8k, 20 prompts, greedy, paired:

| | τ | steady τ |
|---|---|---|
| `-dds -dc 0.4` | 2.80 | 2.91 |
| full block | **3.25** | **3.34** |

19/20 paired wins. By mode: prose 2.48 → 3.03 (n=15), looping 5.98 → 6.05 (n=5, saturated). Marginal verify position costs ~0.5–0.9 ms against a 14.76 ms single-token forward, so cropped positions were nearly free and you lose τ and tok/s both.

**2. Your rejection-sampling read replicates.** Greedy 1.73 → sampled `T=0.8 --min_p 0` 1.88 (+8.7%, 4-prompt cells, not temp 1). Truncation decay confirmed: plain 1.96 vs `top_k 20 top_p 0.9` 1.54 blend.

**3. The overhead is ~4× smaller than unoptimized.** Forced-accept microbench, ms/verify-position: plain 0.184, +top-k 0.395, +top-p 0.379, both 0.538. ~17 launches/position, two host syncs per position. Note this compounds your point — truncation triples the cost while shrinking the gain.

`p(d)`, `q(d)` and the uniform are position-independent (pure functions of the verify logits + stashed proposal); only the reject-path residual sample is lazy. Hoisting them out of the serial loop → one host sync per round:

| ~5 positions | before | after |
|---|---|---|
| plain | ~0.9 ms/round | ~0.25 ms/round |
| top-k + top-p | ~2.7 ms/round | ~0.6 ms/round |

Greedy bit-identical across processes (327/408 accepts), reference parity intact, ~40 lines in one function. Counterintuitive: eager-batching all positions beat lazy-per-position 3.5× (70 vs 251 µs plain, 464 vs 1620 µs combo) — at V=248k, launch count dominates bytes, so redundant V-wide work is cheaper than per-position dispatch.

**4. Coupled Gumbel — two things to check.** It's likely also lossless (shared noise + token-match ⇒ accept requires draft == target's own sample), so losslessness may not distinguish the approaches, which favors yours. And it's a sampled path: we found draft quant quality barely affects greedy acceptance — uncalibrated local 5bpw matched published to 0.01 τ (2.18 vs 2.17; distinct SHA256s, and the one divergent cell's counts differ 182/336 vs 181/344, which same-weights greedy determinism forbids). Mechanism is argmax collapse, which doesn't apply under sampling, so sampled paths should be *more* quant-sensitive. Untested — flagging as a variable. (Uncal 6bpw at 8k: τ 1.80 vs published 5bpw 1.73.)


rktchip:
| Source | What this branch took |
|---|---|
| [#379](https://github.com/turboderp-org/exllamav3/pull/379) — **am423** | The tree: `DFlash2Config` / `DFlash2Model` / `GroupedDynamicCausalConv` / `CandidateSelector` / `propose`, q-aware verify (`p` and `q` from the **same quantized** target `lm_head`), token-match fallback, reference-parity tests |
| [#334](https://github.com/turboderp-org/exllamav3/pull/334) — **RodriMora**, post-review | Triton conv **body** (eager path kept as fallback + test oracle); stock `RMSNorm`; fp32 residual + conv `finish()`; `draft_conf` as `[anchor, scores…]`; tap reorder by `target_layer_ids`; no `export_state_order`; no attention-span patches; no `_DFlash2Norm`; draft length not pinned to 7 |
| **z-lab/dflash** | Algorithm and checkpoint contract (selector walk, q-aware rejection) |
| **MiaAI-Lab/exllamav3** | Convert practice only: raw-fp16 tensors carried through convert via an explicit caps flag (`retain_raw_fp16`), not a fake `maybe_quantize`. Not their server, NVFP4 KV, or MTP launcher |

rktchip:
## Behavior

```
python examples/chat.py -m <target> -dm <dflash2-draft>
```

`-dm` autodetects DFlash2 and runs it on the **same** DFlash v1 iterate / cache / rollback path. No new CLI.

- Greedy walk + token-match by default (same shape as v1)
- Shared plain temp / top-k / top-p across the batch → sampled walk + q-aware rejection, lossless w.r.t. the **quantized** target
- Penalties, min-p, or mixed samplers → greedy + token-match
- `input_embedding_scale` applies to mask/noise columns only; the anchor column is unscaled
- `output_multiplier` / `final_logit_softcapping` applied to draft logits before the selector
- Loud fail (`ValueError` / `NotImplementedError`, never `assert`): TP targets; CFG (`len(sequences) > 1` or live `cfg_scale ≠ 1`); draft cache tokens/pages that can't hold `block_size`. Both are defense-in-depth — on this tree a CFG tuple already dies in `Job` construction and no public `Cache` constructor builds a sub-256-token cache, so these pin the invariant if either relaxes. `--failure-modes` proves both on silicon.

**Ops note, possibly worth a line in serve docs:** Tabby's default `SS_Fused` includes min-p, which is ineligible for q-aware verify under the lossless-only policy, so the default sampler is expected to fall back to greedy there (verified in-harness only — TabbyAPI has no `-dm` path yet, so this is not a Tabby-measured number); the q-aware path needs `--min_p 0` and no repetition penalty. The harness prints a note when the sampler is ineligible.

---

## Tests

```
pytest tests/test_dflash2_*.py tests/test_v1_dflash_regression_.py tests/test_dflash2_accept_rule_.py
# 77 passed, 3 skipped
```

- Skips need real weights via `EXL3_TEST_DFLASH2`. No Hub, no GPU for the default job — synth fixture under `tests/data/dflash2_synth/`, weights gitignored and built on demand by a session fixture.
- **`test_dflash2_accept_rule_.py`** is the file we'd most expect to be useful on any tree: replicate-mean chi-square on synthetic `p`/`q` across four regimes (including `q` near-disjoint from `p`, flatter, and peaked on a low-`p` token), asserting the propose-accept-resample output matches `p` within Monte-Carlo error. Green twice. One implementation note that cost us a debugging round: the chi-square bound has to use the actual live bin count, not the nominal vocab size.
- Same suite with the z-lab `dflash` package import blocked: verify / confidence / Triton / taps / synth all still run; only reference-parity comparisons skip.
- Triton-vs-eager parity: strided-view maxabs 1.9e-3 fp16, 4.8e-7 fp32.
- v1 regression pins: hook identity, no scale key, plain-block load, default path bit-identical.
- Uncalibrated `convert.py` on the fixture: raw keys land fp16, no EXL3 under those prefixes, converted dir loads as `DFlash2Model`. Emission pinned by `test_raw_fp16_convert_contract` and `test_convert_emission_covers_raw_tensors`.

Quality against a published pair is not in CI:

```
PYTHONPATH=. python eval/eval_dflash2_acceptance.py -m <target> -dm <draft>
```

Protocol (cells A–D, τ, context sweep, `--failure-modes`, `--profile`) is in the harness header. Every cell echoes its resolved config (prefix / chunk / crop / new / prompts) and a load-excluded TTFT — added after a crop-vs-chunk confound cost us a round of measurement.

---

## Silicon

**RTX 5090 · Qwen3-27B EXL3 3.5bpw target + DFlash2 EXL3 5.0bpw draft · same seed**

Headline cells, full block, no crop:

| Cell | τ | Accept rate | Steady tok/s | vs A |
|---|---|---|---|---|
| A — no draft | — | — | 64–77 | 1.0× |
| B — in-checkpoint `--mtp` | 1.89 | 0.563 | 158 | ~2.3× |
| **C — `-dm` greedy, full block** | **2.17** | 0.376 | 217 / 161 / 166 | **~2.3–3.1×** |
| C at 8k, 20 prompts | **3.25** | 0.475 | 3.34 steady τ | — |
| D2 — sampled `T=0.8 --min_p 0` | 1.96 | 0.492 | 126–196 | ~2.4× |

Per-round phases (D2, after first touch):

| Phase | Time |
|---|---|
| target verify | 15–19 ms |
| selector walk | ~2–3 ms in-run (~1.1 ms isolated) |
| draft forward | ~2 ms |
| accept loop | ~1 ms (`accept_us` nonzero ⇒ q-aware path ran) |
| `update_kv_from_target` | ~0.4 ms |

**verify ≫ select ≫ draft.** This is why we didn't pursue kernel work: a fused selector walk measured at ~0.5 ms of saving and the Triton launch-config sweep at ~40 µs/round, against a 15–19 ms verify. Whatever your CUDA kernels gain over Triton here is real but below the noise floor of this pair's tok/s.

**700+ accepted draft tokens across the first real rounds, zero crashes.** TP / CFG / page / cache guards held.

### How to read any number above

- **Blend and steady are different columns.** With four prompts, prompt-1 prefill dominates blend; the ×-vs-A claims are steady-state only.
- **Blend τ is total accepts over total rounds**, not the mean of per-cell τ.
- **Same-seed repeats move a single cell's τ by up to ±0.6** (GPU reduction nondeterminism). Single cells are ±0.3 at best; only paired comparisons and blend-level patterns are interpretable.
- **Blends are prompt-mix dependent.** Short factual prompts echo-loop at τ 5–6.5, so any cross-cell claim must be paired same-prompt. All claims here are.
- **Acceptance tracks the predictability of the target's output, not context length.** Our context sweep varied response regime along with length, so no context-length conclusion here is clean — "no collapse at 64k" is kept only as a confounded serving observation.

<details>
<summary>Full cell archive (including superseded and corrected rows)</summary>

| Cell | Setup | τ | Accept | Steady tok/s | Note |
|---|---|---|---|---|---|
| C-crop | greedy `-dds -dc 0.4` | 1.91 | 0.465 | 122–165 | superseded by no-crop |
| C-ts0 | `tap_shift=0` | 1.95 | 0.486 | 153.4 | taps are not the model-card gap |
| C-local | local uncal 5bpw, short ctx | 1.53 | 0.417 | 112.4 blend | first convert |
| C-r5 | local uncal 5bpw reconvert | 1.69 | 0.444 | 152.6 blend | was mislabeled "C-cal" — never calibrated (driver bug) |
| C-r5-8k | same, 8k | 1.55 | 0.394 | 61.7 blend | vs published 1.73 at 8k |
| C-6bpw-8k | local uncal 6bpw, 8k | 1.80 | 0.498 | 98–185 | bits help on any corpus |
| C-r5-nocrop-hl | local uncal 5bpw, full block | 2.18 | 0.378 | 161.9 blend | == published to 0.01, SHA256-distinct weights |
| C-batch | 4 prompts, one `generate` | 1.75 | 0.488 | 99.9 blend | consensus survives batching |
| C-mixed | batch + job-2 greedy sampler | 1.77 | 0.500 | 121.2 blend | fallback safe |
| D3 | `T=0.85 top_k 20 top_p 0.9` | 1.54 blend | 0.459 | 111–156 | truncation path runs; D2 is the plain-sampler evidence |
| C-2k-20 | full block, 20 prompts, 2k | 1.68 | 0.319 | 138.1 blend | weakest at 20 too → real, but mechanism is target regime (no loops at 2k), not draft decay |
| C-2k-r1/r2 | same 2k cell, cold vs warm | 1.36 / 1.42 | — | 82.6 / 123.4 | warmth doesn't move τ |
| C-2k-last | 2k after 32k in-process | 1.46 | 0.442 | 102.7 blend | order irrelevant |
| C-2k-crop | 2k, offset-500 filler window | 1.78 | 0.492 | 100.3 blend | content decides |
| C-nocrop sweep | 2k / 32k / 64k | 1.54 / 2.26 / 1.82 | — | 99.7 / 33.9 / 17.7 blend | τ clean; see reading notes on ctx |
| C-64k-chunk32k | `--chunk_size 32768`, 64k | 1.82 | 0.290 | 116 / 73 / 114 | **no** chunk effect at 64k; accepts bit-identical to 4k chunks (343/840) |
| C-post-perf | greedy, after the launch-batching commit | 1.73 | 0.445 | 166 / 97 / 142 | **327/408 accepts, bit-identical** to pre-perf |
| D-post-perf | sampled, after the same | 1.88 | 0.450 | 131 / 107 / 105 | τ move is stream noise; steady −4 / +17 / +8% |
| C-cal, C-cal-8k | — | — | — | — | SUPERSEDED: were uncalibrated (driver bug); see C-r5 rows |

Two invariants from this table may be useful as regression tooling on any tree: the 4-prompt greedy default returns **327/408** accepts, and 64k greedy returns **343/840** independent of chunk size.

</details>

### Convert status

Two local uncalibrated converts of the real bf16 draft (5bpw, 6bpw). At no-crop the local 5bpw matches the published 5bpw to 0.01 τ (2.18 vs 2.17), per-cell near-identical — with SHA256-distinct weights and resolved paths logged in both runs, so this is not the same file loaded twice. The divergent cell's accept counts differ (182/336 vs 181/344), which same-weights greedy determinism forbids.

So **there is no calibration gap at this magnitude.** The 1.69 → 1.91 delta we originally attributed to calibration corpus was the `-dc` crop discarding the local draft's good positions. Mechanism for the match: at temperature 0 the selector takes an argmax over candidate scores, so quantization noise that never crosses an argmax boundary produces an identical token path. **Scoped claim: draft quantization quality barely affects acceptance at greedy.** Sampled-path sensitivity is untested and should go the other way, since a sampled walk draws from the softmax and noise below the argmax boundary still shifts draw probabilities.

A true calibrated convert fails structurally on this checkpoint (the draft borrows the target embedding; convert attaches no target). **Ship published draft weights regardless.** Key-set diff: 189 tensors match modulo codebook naming; 10 conv `base_kernel`s are BF16 published vs F16 local.

---

## Commits

1. `arch: DFlash2DraftModel with taps, attach guards, and block hook` — includes the one-line v1 `block_cls`
2. `modules: grouped dynamic conv with Triton serve path and selector`
3. `generator: DFlash2 propose branch with q-aware verify, CFG/page guards`
4. `convert+tests: raw-fp16 gate, synth fixture, CPU coverage`
5. `tests+eval+doc: v1 regression pins, harness modes, synthetic losslessness proof, usage notes`
6. `perf: batched accept flags, hoisted walk gathers, single-sync coin`

Commit 6 is separate on purpose: it removed ~200 launches/round (profiler-measured — 17 launches per accept position, 171 per walk round) while leaving greedy output **bit-identical** and strict reference parity intact. Folding it into (3) would hide that. A `searchsorted` substitution in the walk was tried and reverted — the walk is autoregressive, so a stream change cascades into trajectory divergence (2.8e-3 measured) rather than staying at ulp noise. The same substitution is safe on the bonus draw, which is terminal, and shipped there.

---

## Deliberately out of scope

- Mid-job switch onto in-checkpoint MTP
- NVFP4 / extra KV dtypes
- CFG support; tensor-parallel targets
- Fused selector kernel and Triton autotune lock-in (measured too small — see phase table)
- Calibrated convert (structurally blocked above)

---

## One finding outside this branch

Surfaced while validating DFlash2; doesn't belong in this PR, offered as-is.

**Batch-1 decode is launch-bound, not bandwidth-bound.** Single-token target forward, short context, no draft, n=32: wall **14.76 ms** vs summed kernel time **11.69 ms** over **914 launches** → a 3.07 ms host/launch gap, at ~1312 GB/s (~73% of peak). The kernels look healthy; the milliseconds are between them. Separately, round time is 18.71 ms against that 14.76 ms forward, so ~4 ms per round is generator bookkeeping outside the forward. A per-batch-size CUDA-graph capture of the decode step looks like the shape of a fix, with two obstacles we didn't test: capture-safety through EXL3's custom ops, and needing a graph cache plus eviction policy per batch size. Mentioned mainly to explain why we stopped optimizing where we did — DFlash2-side headroom is provably small next to it.

---

## Things we didn't resolve

Listed for completeness, not as asks.

- **Sampled-path quantization sensitivity.** Greedy shows none; the argmax-collapse mechanism above predicts sampled should differ. One cell each against a published draft would settle it.
- **MTP vs DFlash2 crossover.** DFlash2 no-crop beats `--mtp` at batch 1 / short context here (161–217 vs 158 steady, τ 2.17 vs 1.89). But MTP's drafting overhead is ~0, DFlash2's is a context-independent ~3 ms of walk + accept, and q-aware verify needs a *homogeneous* plain sampler across the batch — so heterogeneous serving batches fall back to greedy while MTP keeps its 0.563 accept rate. We have no batch-diversity data, so we deliberately print no decision rule.
- **Long-context cost model.** Marginal verify cost per position rises with KV, so full-block should stop dominating somewhere. Our long-context timings weren't clean enough to locate it; τ held at every length we measured.

PR #383, opened 2026-09-19
Fix pre-Ampere miscompilation and silent-corruption hazards in the deterministic GEMM paths

Frenchy2k1:
## Summary

Four fixes for the v1.5.0 deterministic-GEMM and tiled-mix code paths, discovered while building the extension for pre-Ampere hardware (V100, sm_70). Upstream's CI only compiles sm_80+ arches, so these hazards are invisible there — but any multi-arch build (or a pre-Ampere runtime) hits them.

### 1. Arch-guard the sm_80+ inline asm (det_gemm.cuh, hgemm_f16acc.cu)

The deterministic-GEMM wrappers (`det_cp_async16`, `det_mma_s8`, `det_cp_async_commit/wait`, `det_ldmatrix_x4`) and hgemm_f16acc's local wrappers (`cp_async16`, `mma_f16`, `mma_f32`, `ldmatrix_x4[_trans]`) wrap sm_80+ instructions (cp.async, mma.m16n8k16/m16n8k32) in `__device__` functions with no `__CUDA_ARCH__` guard. Consequences:

- **Build failure**: ptxas rejects the TU outright when the arch list includes anything below sm_80 (`error: Feature 'cp.async' requires .target sm_80 or higher`).
- Fix: `#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800` guards with no-op stubs (ldmatrix at 750). The wrappers stay defined on every arch, so kernel bodies compile everywhere; ptxas never sees the instructions below sm_80.

### 2. Gate the tiled GR-mix path on cc >= 8 (hyperconnections.py)

`self.tiled` selected the int8 tensor-core path on shape constraints alone. On pre-Ampere the kernels would launch with arch-guarded no-op compute stages and silently produce garbage. Added `ext.g_get_cc(dev.index or 0) >= 8` to the gate; the cuBLAS fallback serves pre-Ampere.

### 3. Gate the deterministic router GEMM on cc >= 8 (routing_gemm.cu)

`routing_gemm_det_fits` accepted shapes the kernels cannot serve below sm_80: the kernels use cp.async and mma.m16n8k32 s8 (both sm_80+) and need 97 KB dynamic smem — over the 96 KB pre-Ampere ceiling. On V100 the launch failed with a GPU assert (`invalid argument`). Added an arch check to the fits predicate; the cuBLAS router path serves those devices.

### 4. Scope the determinism claims to sm_80+ (routing.cu, routing_gemm.cu, hyperconnections.py)

The comments claimed rank-identical routing/mixing on "ANY architecture". With the gates above, the identical-streams guarantee holds only where all ranks take the int8 path (sm_80+); pre-Ampere devices take the cuBLAS path, which is device-dependent — though still uniform within a single-arch fleet. The docs now say so.

## Testing

- Extension build with `TORCH_CUDA_ARCH_LIST="8.0;8.6;8.9;9.0"` (the CI arch list) — succeeds; sm_80+ codegen is byte-identical to the unguarded build since the guards only exclude compilation below sm_80.
- Functional runs on 4x V100 (sm_70): model load + generation clean, the deterministic paths correctly fall back to cuBLAS, no GPU asserts.
- sm_80+ behavior unchanged: the guards compile identically on Ampere+ passes, and the runtime gates only ever exclude pre-Ampere.

## Scope note

These fixes cover the deterministic-GEMM wrappers, the tiled GR-mix gate, and the router GEMM. A full pre-Ampere tree build additionally requires guards on the comp-unit kernels (`exl3_gemm_inner.cuh` carries unguarded cp.async that reaches the `exl3_comp_unit_*` TUs) — that is follow-up work, out of scope here.


Frenchy2k1:
Note: this is a pre-patch before opening a PR to add support for sm70.

Issue #390, opened 2026-09-21
DFlash2 draft loader rejects checkpoints with .weight-suffixed codebook keys


rktchip:
# DFlash2 draft loader rejects checkpoints with `.weight`-suffixed codebook keys

## Symptom

Loading a DFlash2 draft checkpoint whose candidate-selector tensors carry the
older `.weight`-suffixed names fails at load:

- `candidate_selector.predecessor_codebook.weight`
- `candidate_selector.successor_codebook.weight`

`CandidateSelector.load` (`exllamav3/modules/arch_specific/dflash2.py`) looks
up the bare keys only (`optional=False`), so the load raises a missing-tensor
error even though the weights are present under the suffixed names.

## Root cause

Key-convention split between trainer generations: older trainers append
`.weight` to the codebook keys, 1.5.1's loader demands the bare keys. (1.5.1's
own `get_tensors` writes the bare keys, so newly converted checkpoints are
unaffected — this only bites older/shared checkpoints.)

## Suggested fix

Bare-first with suffixed fallback, per codebook (~6 lines):

```python
self.pred_codebook = self.config.stc.get_tensor(
    self.key_pred, self.device, optional=True, allow_bf16=True)
if self.pred_codebook is None:
    self.pred_codebook = self.config.stc.get_tensor(
        self.key_pred + ".weight", self.device, optional=False,
        allow_bf16=True)
# ... same for succ_codebook
```

## Verification (local, Windows, RTX 5090)

With the fallback applied, a 4.00bpw DFlash2 draft (suffixed keys) loads
against a 3.5bpw EXL3 target and serves: ~201 tok/s decode on 32k-context
code generation, token-identical to the AR baseline on both seeds, plus
further gains with lookup-first drafting (~293 tok/s, still bit-exact).
Without it, the same checkpoint fails to load at all.

Happy to PR the fallback if wanted.

turboderp:
PR is merged into `dev` now

PR #391, opened 2026-09-21
dflash2: accept .weight-suffixed codebook keys (older trainers)

rktchip:
Follows up #390.

Older DFlash2 trainers append `.weight` to the candidate-selector codebook keys (`candidate_selector.predecessor_codebook.weight` etc.). `CandidateSelector.load` looks up the bare keys with `optional=False`, so those checkpoints fail to load even though the weights are present.

This tries the bare key first (`optional=True`) and falls back to the suffixed key. Same bytes either way; newly converted checkpoints (bare keys) take the first path unchanged.

Verified locally on Windows/RTX 5090: a 4.00bpw suffixed draft loads against a 3.5bpw EXL3 target and serves — ~201 tok/s decode on 32k code, token-identical to the AR baseline on both seeds.


turboderp:
:+1: thx

PR #392, opened 2026-09-21
CPU MoE: streamed single-row decode from the pinned arena (EXL3_MOE_STREAM_DECODE)

lmganon16:
## Summary
Experimental, opt-in. With whole-layer CPU offload (`-mcl`) every routed expert of a decode step is computed by the worker while the GPU waits; on a desktop CPU that wait is most of the decode budget. With the pinned arena (#341) the parent already has every expert's contiguous `[gate | up | down]` block mapped into the device address space, so at single-row steps this gathers the selected experts' blocks over PCIe into a VRAM staging buffer with one kernel and runs the resident fused MoE kernel (`exl3_moe`) over them through pointer tables — the streamed prefill tier's scheme without its per-batch host loop. `EXL3_MOE_STREAM_DECODE_CPU=k` leaves the k lowest-ranked picks per layer with the worker, so the link and the cores work at the same time. No host synchronization anywhere on the path.

Usage: `EXL3_MOE_PINNED_ARENA=1 EXL3_MOE_STREAM_DECODE=1 EXL3_MOE_STREAM_DECODE_CPU=4 <load with -mcl N>`. Sweep k for the host (best where the link and the cores finish together); `EXL3_MOE_STREAM_DECODE=check` validates a new model or host against the worker.

## Results

RTX 3060 12 GB + i5-12400F (6 cores, AVX2 only), 64 GB DDR4-3200, PCIe 4.0 x16 (24 GB/s pinned→device), `-mcl 999 -mct 6`, `EXL3_MOE_PINNED_ARENA=1 EXL3_MOE_STREAM_T=1` in every run (so *off* is the same setup with only this path disabled), 300-token greedy completions, two runs per configuration with the path off at both ends, one model load (a small sweep script that swaps the path in and out of the running host; not part of this PR, I'll publish it separately). tok/s:

| model | off | k=0 | k=1 | k=2 | k=3 | k=4 | k=5 |
|---|---|---|---|---|---|---|---|
| Qwen3.8-Flash-Next 3.05 bpw (512 experts, top-10, 48 offloaded layers) | 11.2 | 14.5 | – | 15.9 | 16.9 | **19.0** | 17.3 |
| lfm2.5-8b-a1b 3.10 bpw (32 experts, top-4, 22 offloaded layers) | 30.7 | 43.2 | **55.5** | 52.3 | 38.7 | – | – |

Prefill is unchanged (the path only claims 1-row steps; the bench's prefill column is flat). Timing at k=4 on Qwen (`EXL3_MOE_STREAM_DECODE_PROF=1`): 22.7 ms gather + 6.6 ms fused kernel + 1.8 ms worker wait per token, out of ~53 ms.

## Numerics

`EXL3_MOE_STREAM_DECODE=check` (both paths every layer call, worker result returned): Qwen 2064 layer calls, max rel 4.6e-2, typical 1e-2, non-finite 0; lfm2.5 946 layer calls, max rel 4.7e-2, non-finite 0. Same fp16-intermediates-vs-fp32-worker difference as the streamed prefill tier's against the CPU tail.

## What's in it

- `exllamav3_ext/moe_stream_gather.cu`: the gather kernel (block copy + aux pointer gather in one launch), bound as `moe_stream_gather`. Hits the link rate (25.1 GB/s here).
- `model/moe_stream_decode.py`: per-device state (arena aliases, per-layer tables, static fused-kernel descriptors), the forward, `check` and `PROF`.
- `model/moe_cpu_host.py` (~40 lines): knobs in `MoeCpuTuning`, state built in `ensure_started`, the `rows == 1` branch at the top of `submit_prefill`, staging accounted for in `_device_buffers`, torn down first in `shutdown`. `block_sparse_mlp_cpu.py` is untouched.
- `tests/test_moe_stream_decode_.py`: kernel vs pinned host memory (two chunks, -1 slots, aux table) and the swizzled-arena round trip through `moe_unswizzle_trellis` (K1/K3/K8) — the AVX-512 host case, which I cannot run here.
- `doc/env_vars.md`: three entries.

## Behaviour and limits

- Per-layer eligibility is `_stream_fused_t` (fused-kernel-eligible layers only); rows ≥ 2, ineligible layers and everything without the pinned arena take the worker path.
- Swizzled arenas (AVX-512 tiers) are restored to native order in VRAM (`moe_unswizzle_trellis`), one launch per projection.
- Any error disables the path for the run with a traceback; the worker serves everything.
- Staging is top-k × expert block per device (17 MiB on Qwen), allocated with the persistent device buffers so autosplit sees it.
- Untested: multiple GPUs, draft/MTP components (state is per host, so it should follow), gateless experts, Windows.
- k is manual for now. Deriving it from the bandwidth probe plus a CPU-rate probe (the way `stream_t` is derived) is the natural follow-up; so is `submit_issue_fused` for the worker share instead of the copy path.

Developed and measured on the hardware above with Claude's assistance; every configuration in the tables was run here.


lmganon16:
i uploaded a sweep script for finding best `k` on your pc: https://github.com/lmganon16/exl3-stream-decode-bench
people can do it manually too, but for bigger models the loading time is too long so i've been using this script instead.
for example:
## lfm-8b-a1b-mul1-3.10bpw
- 22 offloaded layers, 32 experts, top-4, 3.94 MiB per expert block: 363 MB of routed expert weights per token (14.9 ms over the link with every pick streamed)
- GPU: NVIDIA GeForce RTX 3060 12 GiB on gen4 x16, pinned->device 24.4 GB/s
- CPU: 12th Gen Intel(R) Core(TM) i5-12400F, 6 worker threads (avx2, native layout), 62 GiB RAM
- exllamav3 1.5.1 (/home/john/TND/exllamav3-up/exllamav3), -mcl 999, 300 tokens per run, 2 runs per config

| config | gen t/s | vs off | prefill t/s | calls |
|---|---|---|---|---|
| off | 29.72 | - | 1345 | 0 |
| k0 | 45.20 | +52% | 1435 | 6600 |
| k1 | 56.93 | +92% | 1405 | 6600 |
| k2 | 49.04 | +65% | 1466 | 6600 |
| k3 | 37.94 | +28% | 1448 | 6600 |

best: k1 at 56.93 t/s, +92% over off (29.72 t/s); streams 272 MB per token (11.2 ms at 24.4 GB/s)

treo:
I gave this a try with Qwen 3.8 Flash Next 2.05bpw.

My system is an  RTX3090, Ryzen 9 5950X, 128GB DDR4 @ 3200MT/s.

Running the following over a sweep from k=0..9
```
EXL3_MOE_STREAM_DECODE=1 EXL3_MOE_STREAM_DECODE_CPU=$k EXL3_MOE_STREAM_T=1 EXL3_MOE_PINNED_ARENA=1 python ../exl3-arena-stream/eval/perf.py --model_dir ../../tabbyAPI/models/Qwen3.8-Flash-Next-exl3-2.05bpw_h4_ng4/ -mcl 48 -ngr -mct 15 -spf --max_length 2048
```

With that I had 12GB VRAM usage.

| k (CPU experts) | GPU-streamed / CPU-computed | Median t/s | Min | Max |
|---|---|---|---|---|
| 0 | 10 / 0 | **27.17** | 26.96 | 27.27 |
| 1 | 9 / 1 | **26.83** | 25.81 | 27.01 |
| 2 | 8 / 2 | **27.71** | 26.79 | 28.02 |
| 3 | 7 / 3 | **28.83** | 28.70 | 28.87 |
| 4 | 6 / 4 | **29.74** | 25.07 | 29.96 |
| 5 | 5 / 5 | **31.24** | 31.08 | 31.37 |
| 6 | 4 / 6 | **32.77** | 32.70 | 33.09 |
| 7 | 3 / 7 | **31.65** | 31.54 | 32.09 |
| 8 | 2 / 8 | **30.54** | 30.24 | 30.84 |
| 9 | 1 / 9 | **29.93** | 29.20 | 29.97 |

As the 3090 is capable of loading more layers, I also did a sweep with that, but k=6 was the fastest there too.
```
EXL3_MOE_STREAM_DECODE=1 EXL3_MOE_STREAM_DECODE_CPU=$k EXL3_MOE_STREAM_T=1 EXL3_MOE_PINNED_ARENA=1 python ../exl3-arena-stream/eval/perf.py --model_dir ../../tabbyAPI/models/Qwen3.8-Flash-Next-exl3-2.05bpw_h4_ng4/ -mcl 38 -ngr -mct 15 -spf --max_length 512 --cache_size 262144 --chunk_size 2048
```

```
 -- CPU MoE worker started: 38 layers, avx2, 15 threads
 -- Warming up...
 -- stream-decode: cuda:0, 23 arena chunks mapped, 4 experts per layer streamed, 6 computed by the worker, staging 11 MiB
 -- Loading tokenizer...
 -- Bitrate: 2.09 bpw / 4.01 bpw (head)
 -- Chunk size: 2048

Generation
Context      0: S=1      36.43 tokens/s
Context    256: S=1      36.66 tokens/s
```

For comparison sake, my normal setup is with expert offloading:
```
EXL3_MOE_PINNED_ARENA=1 python ../exl3-arena-stream/eval/perf.py --model_dir ../../tabbyAPI/models/Qwen3.8-Flash-Next-exl3-2.05bpw_h4_ng4/ -mcs 395 -ngr -mct 15 -spf --max_length 256 --cache_size 262144 --chunk_size 2048
```

```
Generation
Context      0: S=1      37.05 tokens/s
Context    256: S=1      37.04 tokens/s
```

So the results seem to be strictly worse, unless I'm using this wrong.

Ph0rk0z:
Is K the number of cores used? This didn't beat expert streaming offload. The good news I guess is that multi-gpu didn't hurt it. And EXL3_MOE_STREAM_T=1 is required? I don't see it anywhere in the PR. This PR is around 25-26t/s and expert streaming is 28-29t/s on GLM flash 3bpw. Am using the 512_bw kernel though since I have xeon.

PR #393, opened 2026-09-21
aarch64: guard x86 AVX intrinsics so the extension builds on Grace/GB10/Jetson

vcruz305:
On aarch64 (NVIDIA GB10 / DGX Spark, Grace, Jetson) the extension does not build: several CPU paths include `<immintrin.h>` and use AVX2/AVX-512 intrinsics unconditionally, and `all_reduce_cpu.cu` requires AVX2 outright.

This guards those paths behind `__aarch64__` / `__arm__` so the same sources compile on both, with no functional change on x86.

## What it does

- `avx2_target.{cpp,h}`, `avx512_target.{cpp,h}`: compile the vector bodies only where the intrinsics exist.
- `parallel/all_reduce_cpu.cu`, `all_reduce_cpu_avx2.cpp`, `all_reduce_cpu_avx512.cpp`: guard the AVX reduce kernels, so the native TP CPU reduce degrades rather than failing to compile.
- `cpu/moe_mul1.cpp`, `cpu/moe_handoff.cu`: provide a scalar fallback for the CPU MoE offload path.
- Spin hint: `asm volatile("yield")` in place of `_mm_pause()` on ARM.

Every change is inside a preprocessor guard, so x86 codegen is unchanged. 9 files, +56/-7.

## Testing

Built and run on GB10 (aarch64, sm_121, CUDA 13.0) with Python 3.12 and torch 2.9, where the extension previously would not compile. Inference and tensor-parallel load both work on that machine.

Not tested on Jetson or Grace specifically, and I have not verified an x86 build from this exact branch, since every hunk is guarded and x86 takes the pre-existing path. Happy to adjust the guard style if you would rather key off a CMake/setup feature check than the compiler macros.


turboderp:
:+1: thx

I have no way to test this at the moment, but it seems fine.

